PR #9, opened 2019-01-03
App secrets component

Dhouti:


pkg/controller/summon/components/app_secrets.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	fetchTarget := &corev1.Secret{}
+	err = ctx.Get(ctx.Context, types.NamespacedName{Name: fmt.Sprintf("summon.%s.app-secrets", instance.Name), Namespace: instance.Namespace}, fetchTarget)
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+	}
+
```
coderanger: This needs a call to SetControllerReference somewhere, probably inside CreateOrUpdate like we do for pullsecrets.

pkg/controller/summon/components/app_secrets.go
```diff
@@ -0,0 +1,106 @@
...
+	_, err = controllerutil.CreateOrUpdate(ctx.Context, ctx, newSecret, func(fetchTarget runtime.Object) error {
+		existing := fetchTarget.(*corev1.Secret)
+		// Sync important fields.
+		existing.ObjectMeta = newSecret.ObjectMeta
+		existing.Type = newSecret.Type
+		existing.StringData = newSecret.StringData
```
coderanger: Use `.Data`, `.StringData` is an input-only field.

pkg/controller/summon/components/app_secrets.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		return reconcile.Result{Requeue: true}, errors.Wrapf(err, "app_secrets: Failed to get existing app secrets")
+	}
+
+	postgresSecret := &corev1.Secret{}
+	err = ctx.Get(ctx.Context, types.NamespacedName{Name: fmt.Sprintf("summon.%s-database.credentials", instance.Name), Namespace: instance.Namespace}, postgresSecret)
+	if err != nil {
+		return reconcile.Result{Requeue: true}, errors.Wrapf(err, "app_secrets: Postgres password not found")
+	}
+	postgresPassword := postgresSecret.Data["password"]
```
coderanger: Probably put an error check on here in case of a malformed (probably empty) secret, that's the `postgresPassword, ok := ..` syntax where `ok` will tell you if the key was present.

pkg/controller/summon/templates/celerybeat/statefulset.yml.tpl
```diff
@@ -31,11 +31,6 @@ spec:
       - name: init
```
coderanger: This whole thing can go away now.

pkg/controller/summon/components/app_secrets.go
```diff
@@ -64,7 +70,10 @@ func (comp *appSecretComponent) Reconcile(ctx *components.ComponentContext) (rec
 	if err != nil {
 		return reconcile.Result{Requeue: true}, errors.Wrapf(err, "app_secrets: Postgres password not found")
 	}
-	postgresPassword := postgresSecret.Data["password"]
+	postgresPassword, ok := postgresSecret.Data["password"]
+	if !ok {
+		return reconcile.Result{}, errors.Wrapf(err, "app_secrets: Postgres password not found in secret")
```
coderanger: There is no error to wrap.

coderanger: `errors.New("message")` or `errors.Errorf("message %s", val)`

pkg/controller/summon/components/app_secrets.go
```diff
@@ -92,9 +101,13 @@ func (comp *appSecretComponent) Reconcile(ctx *components.ComponentContext) (rec
 	_, err = controllerutil.CreateOrUpdate(ctx.Context, ctx, newSecret, func(fetchTarget runtime.Object) error {
 		existing := fetchTarget.(*corev1.Secret)
 		// Sync important fields.
+		err := controllerutil.SetControllerReference(instance, existing, ctx.Scheme)
+		if err != nil {
+			return err
```
coderanger: Should wrap this error.

pkg/controller/summon/templates/celerybeat/statefulset.yml.tpl
```diff
@@ -27,40 +27,6 @@ spec:
...
-        volumeMounts:
-        - name: config-volume
-          mountPath: /etc/config
-        - name: secrets-orig
-          mountPath: /etc/secrets-orig
-        - name: app-secrets
-          mountPath: /etc/secrets
-        - name: postgres-credentials
-          mountPath: /postgres-credentials
-      - name: volumeperms
```
coderanger: Little bit too much, this is still needed.

PR #10, opened 2019-01-03
[IMP] add custom template functions

blaggacao:
Some templates might just cry for a strings.join helper.
This implements `StringsJoin` custom template function.

Not sure if modifying the templating API is desirable for this project, if so, `strings.join` migtht be the first and foremost candidate.

Use: `{{ StringsJoin .Instance.Spec.YourArray ", " }}`

### DIFF
--- a/pkg/templates/templates.go
+++ b/pkg/templates/templates.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"net/http"
 	"path"
+	"strings"
 	"text/template"
 
 	// "github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


blaggacao:
To avoid this hack: `{{ if .Instance.Spec.Modules }}, "--modules", {{ range $index, $element := .Instance.Spec.Modules }}{{ if $index }},{{ end }}{{ $element }}{{ end }}{{ end }}`

coderanger:
What do you think about just jumping all the way to Sprig instead? It has a `join` function: https://masterminds.github.io/sprig/string_slice.html 

And it's what Helm uses so it would be more familiar for people coming from that.

blaggacao:
Ah, all the better. :smile: So I'll close this.

PR #12, opened 2019-01-07
Rotate fernet

Dhouti:


pkg/apis/summon/v1beta1/summonplatform_types.go
```diff
@@ -83,6 +85,9 @@ type SummonPlatformSpec struct {
 	// Slack API Key Secret Definition
 	// +optional
 	NotificationSecretRef NotificationSecretRef `json:"secretRef,omitempty"`
+	// Fernet Key Rotation Time Setting
+	// +optional
+	FernetKeyRotation time.Duration `json:"fernetKeyRotation,omitempty"`
```
coderanger: Change name to something like lifetime or lifespan.

pkg/controller/summon/components/app_secrets.go
```diff
@@ -35,6 +37,25 @@ import (
 
 type appSecretComponent struct{}
 
+type fernetKeyEntry struct {
+	Data []byte
```
coderanger: Maybe call this `Key` so the two fields aren't named so similarly.

pkg/controller/summon/components/app_secrets.go
```diff
@@ -73,10 +94,25 @@ func (comp *appSecretComponent) Reconcile(ctx *components.ComponentContext) (rec
...
+	formattedFernetKeys, err := comp.formatFernetKeys(fernetKeys.Data)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
 	appSecrets.Data["DATABASE_URL"] = []byte(fmt.Sprintf("postgis://summon:%s@%s-database/summon", postgresPassword, instance.Name))
 	appSecrets.Data["OUTBOUNDSMS_URL"] = []byte(fmt.Sprintf("https://%s.prod.ridecell.io/outbound-sms", instance.Name))
 	appSecrets.Data["SMS_WEBHOOK_URL"] = []byte(fmt.Sprintf("https://%s.ridecell.us/sms/receive/", instance.Name))
 	appSecrets.Data["CELERY_BROKER_URL"] = []byte(fmt.Sprintf("redis://%s-redis/2", instance.Name))
+	appSecrets.Data["FERNET_KEYS"] = formattedFernetKeys
```
coderanger: Should be a []string rather than YAML encoded.

pkg/controller/summon/components/app_secrets_test.go
```diff
@@ -32,6 +33,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const customTimeLayout = "2006-01-02T15-04-05"
```
coderanger: If you change the const in the component to be `CustomTimeLayout`, you can import and use it directly.

pkg/controller/summon/components/defaults.go
```diff
@@ -70,6 +71,14 @@ func (comp *defaultsComponent) Reconcile(ctx *components.ComponentContext) (reco
 		instance.Spec.PullSecret = "pull-secret"
 	}
 
+	// Fernet key rotation default
+	zeroSeconds, _ := time.ParseDuration("0s")
```
coderanger: Move `zeroSeconds` out to a const.

pkg/controller/summon/components/rotate_fernet.go
```diff
@@ -0,0 +1,129 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	summonv1beta1 "github.com/Ridecell/ridecell-operator/pkg/apis/summon/v1beta1"
+	"github.com/Ridecell/ridecell-operator/pkg/components"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+const customTimeLayout = "2006-01-02T15-04-05"
```
coderanger: "2006-01-02T15-04-05Z" would be nice if possible for more https://en.wikipedia.org/wiki/ISO_8601 compat, even if we can't use colons. Also probably include a comment about why this is bastardized 8601 :)

pkg/controller/summon/components/rotate_fernet.go
```diff
@@ -0,0 +1,129 @@
...
+	instance := ctx.Top.(*summonv1beta1.SummonPlatform)
+
+	fernetSecret := &corev1.Secret{}
+	err := ctx.Get(ctx.Context, types.NamespacedName{Name: fmt.Sprintf("%s.fernet-keys", instance.Name), Namespace: instance.Namespace}, fernetSecret)
+	// If secret doesn't exist reconcile to create new
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return true
+		}
+		return false
```
coderanger: This would mean that an error getting the secret would be silently dropped, which is not good.

pkg/controller/summon/components/defaults.go
```diff
@@ -27,6 +27,11 @@ import (
 	"github.com/Ridecell/ridecell-operator/pkg/components"
 )
 
+const fernetKeysLifespan = "8760h"
```
coderanger: Maybe put a "Default" in that name so it's clearer.

PR #13, opened 2019-01-08
Implement accumulator pattern

blaggacao:
_Note: This is more a "Note to myself" and a "Note to @coderanger" rather than a true intent for requesting acceptance. Just close it and have it in this repo's PR inventory in case you may find it useful later on._

A GathererComponent implements WatchPredicateFuncs()
IsGatherable acts upon a *handler.MapObject to determin if Top object
should be reconciled.

It uses the same WatchTypes as used for the OwnerReference, so WatchPredicateFuncs
should take care to avoid reimplementing a OnwerReference semantic.

### DIFF
--- a/pkg/components/reconciler.go
+++ b/pkg/components/reconciler.go
@@ -24,8 +24,10 @@ import (
 
 	"github.com/golang/glog"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


blaggacao:
I guess that's a better approach: https://github.com/Ridecell/ridecell-operator/pull/25/commits/5d5685902354c1c3bd35774ec7c46598926a2408

PR #14, opened 2019-01-08
secret_key component

Dhouti:


pkg/controller/summon/components/app_secrets.go
```diff
@@ -108,13 +108,24 @@ func (comp *appSecretComponent) Reconcile(ctx *components.ComponentContext) (rec
 		return reconcile.Result{}, err
 	}
 
+	secretKey := &corev1.Secret{}
+	err = ctx.Get(ctx.Context, types.NamespacedName{Name: fmt.Sprintf("%s.secret-key", instance.Name), Namespace: instance.Namespace}, secretKey)
+	if err != nil {
+		return reconcile.Result{Requeue: true}, errors.Wrapf(err, "app_secrets: Unable to get SECRET_KEY")
+	}
+	val, ok := secretKey.Data["SECRET_KEY"]
+	if !ok {
```
coderanger: Can we do `!ok || len(val) == 0` to catch the case of it being set to `""`? Probably overkill, but low cost overkill :)

PR #16, opened 2019-01-09
Implement the ready status

coderanger:
A simple feature that maybe got a bit out of hand. Sigh.


coderanger:
So this started as a relatively small component, but that revealed problems in how we manage status updates and fixing them required adding a new set of tools for dealing with statuses. Not sure I 100% love this API but the basic idea is that rather than changing status stuff in-line, you change it by returning a function. This allows replaying a status update against a fresh copy of the object for situations where we have a write collision.

This could maybe go back to a simpler API when serverside apply is available?

PR #24, opened 2019-01-25
Aws summon components

Dhouti:


pkg/apis/summon/v1beta1/summonplatform_types.go
```diff
@@ -96,6 +96,9 @@ type SummonPlatformSpec struct {
 	// SQS queue setting
 	// +optional
 	SQSQueue string `json:"sqsQueue,omitempty"`
+	// AWS Bucket Name
+	// +optional
+	BucketName string `json:"bucketName,omitempty"`
```
coderanger: Any particular reason to make this configurable? Tempted to call a YAGNI on it.

pkg/controller/summon/templates/aws/iamuser.yml.tpl
```diff
@@ -1,10 +1,10 @@
 kind: IAMUser
 apiVersion: aws.ridecell.io/v1beta1
 metadata:
- name: {{ .Instance.Name }}
+ name: {{ .Instance.Name }}-iamuser
```
coderanger: Because things are scoped by type, we don't really need this. Up to you though.

PR #31, opened 2019-01-29
RabbitMQ Vhost CRD and controller.

mehtapratikd:


pkg/controller/rabbitmq_vhost/components/defaults.go
```diff
@@ -0,0 +1,61 @@
...
+		// Use "guest" as the default key.
+		instance.Spec.Connection.Password = "guest"
+	}
+	if instance.Spec.Connection.Username == "" {
+		// Use "guest" as the default username.
+		instance.Spec.Connection.Username = "guest"
+	}
+	if instance.Spec.Connection.Host == "" {
+		// Use "guest" as the default username.
+		instance.Spec.Connection.Host = "http://127.0.0.1:15672"
```
coderanger: The comment here needs updating. And are you sure this is a safe default value?

pkg/apis/db/v1beta1/types.go
```diff
@@ -33,3 +33,9 @@ type PostgresConnection struct {
 	PasswordSecretRef helpers.SecretRef `json:"passwordSecretRef"`
 	Database          string            `json:"database,omitempty"`
 }
+
+type RabbitmqConnection struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
```
coderanger: So rather than storing the password directly in the spec, which would effectively put it in plain text in a bunch of places, we generally want to use a reference to a secret instead. You can use the `SecretRef` type for this purpose :) There is even a handy utility method on it to resolve the actual password, so it should be a small change.

That keeps the actual password in a Secret object, while the RabbitmqVhost just references it.

pkg/controller/rabbitmq_vhost/components/vhost_test.go
```diff
@@ -0,0 +1,102 @@
...
+		instance.Spec.Connection.Password = ""
+		instance.Spec.Connection.Username = ""
+		instance.Spec.Connection.Host = ""
+		fakeFunc := func(uri string, user string, pass string, t *http.Transport) (rmqvcomponents.RabbitMQManager, error) {
+			var mgr rmqvcomponents.RabbitMQManager
+			mgr = &fakeRabbitClient{}
+			return mgr, nil
+		}
+		comp.InjectFakeNewTLSClient(fakeFunc)
+		_, err := comp.Reconcile(ctx)
```
coderanger: You can use the newer ReconcileContext helper method.

pkg/controller/rabbitmq_vhost/components/vhost_test.go
```diff
@@ -0,0 +1,102 @@
...
+		Expect(err).NotTo(HaveOccurred())
+	})
+	It("Create new vhost if it does not exist", func() {
+		comp := rmqvcomponents.NewVhost()
+    mgr := &fakeRabbitClient{}
```
coderanger: gofmt again.

mehtapratikd:
Changes Complete !

config/crds/db_v1beta1_rabbitmqvhost.yaml
```diff
@@ -0,0 +1,77 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
coderanger: File should get removed, we are no longer checking them in to git to reduce diff spam.

helm/templates/crds/db_v1beta1_rabbitmqvhost.yaml
```diff
@@ -0,0 +1,77 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
coderanger: Ditto.

pkg/apis/db/v1beta1/types.go
```diff
@@ -33,3 +33,10 @@ type PostgresConnection struct {
 	PasswordSecretRef helpers.SecretRef `json:"passwordSecretRef"`
 	Database          string            `json:"database,omitempty"`
 }
+
+type RabbitmqConnection struct {
+	Username     string            `json:"username"`
+	Password     helpers.SecretRef `json:"passwordSecretRef"`
+	Host         string            `json:"host"`
+	InsecureSkip bool              `json:"InsecureSkip"` // skip server certificate trust, by default false
```
coderanger: You'll want to mark this `insecureSkip,omitempty`. The first part to match normal Kubernetes naming with camel case, the second bit to mark the field as optional. For whatever reason `omitempty` is what the CRD generator uses to decide is a field should be optional.

pkg/controller/rabbitmq_vhost/components/defaults.go
```diff
@@ -0,0 +1,60 @@
...
+	instance := ctx.Top.(*dbv1beta1.RabbitmqVhost)
+
+	// Fill in defaults.
+	if instance.Spec.VhostName == "" {
+		// Default extension name is just the name of the resource.
+		instance.Spec.VhostName = instance.Name
+	}
+	if instance.Spec.Connection.Password.Key == "" {
+		// Use "password" as the default key.
+		instance.Spec.Connection.Password.Key = "password"
```
coderanger: This is handled by the `Resolve` helper, don't need it here too.

pkg/controller/rabbitmq_vhost/components/defaults.go
coderanger: I still don't love having this as a default value.

pkg/apis/db/v1beta1/types.go
mehtapratikd: Ahh Yes 👍

PR #32, opened 2019-01-29
Move DjangoUser from summon.ridecell.io to db.ridecell.io.

coderanger:
It's more database related and we use Django for plenty of other services where this might be useful too.

Though we would need to deal with the Summon-specific tables if we did.

I could go either way on this one. It is actually summon-specific at the moment because the profile tables are (I think) only used in summon?

### DIFF
--- a/pkg/apis/db/v1beta1/djangouser_types.go
+++ b/pkg/apis/db/v1beta1/djangouser_types.go
@@ -20,24 +20,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-type SecretRef struct {
-	Name string `json:"name"`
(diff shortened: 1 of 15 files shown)


coderanger:
The more I think about this, the more I think we should probably leave it in summon. given that I'm not sure how this would be made generic without a weird mode flag? I'm conflicted because moving it later would be more annoying but also YAGNI.

coderanger:
@jolene-engo The other places to use it would presumably be in microservices, so do you have any opinions? :)

coderanger:
@jengo The other places to use it would presumably be in microservices, so do you have any opinions? :)

coderanger:
Going to close this out, I think it should stay put.

PR #36, opened 2019-01-30
Added Permissions boundaries to iamuser

Dhouti:


pkg/controller/summon/components/iamuser.go
```diff
@@ -43,8 +45,14 @@ func (_ *iamUserComponent) IsReconcilable(_ *components.ComponentContext) bool {
 }
 
 func (comp *iamUserComponent) Reconcile(ctx *components.ComponentContext) (components.Result, error) {
+
+	permissionsBoundaryArn := os.Getenv("PERMISSIONS_BOUNDARY_ARN")
```
coderanger: Add an `if == ""` error check.

PR #37, opened 2019-02-04
Implementing shared databases

Dhouti:
Edit postgres component to use PostgresOperatorDatabase when in need of a shared databased


pkg/apis/summon/v1beta1/summonplatform_types.go
```diff
@@ -38,6 +38,14 @@ type NotificationsSpec struct {
 	SlackChannel string `json:"slackChannel,omitempty"`
 }
 
+// DatabaseSpec is used to specify whether we are using a shared database or not.
+type Database struct {
```
coderanger: This still needs fixing.

PR #45, opened 2019-02-09
Added aws credentials to app_secrets

Dhouti:


coderanger:
Coverage can't be helped, unit tests can't generate those kinds of errors easily.

PR #49, opened 2019-02-10
🚧 Experimenting with a better approach for errors that shouldn't notify.

coderanger:
Still not sure this approach will work, but it's a start maybe? The issue would be that the next converge would see it in Status Error potentially. Or maybe not? Haven't thought it through enough but it's Saturday and I'm tired so putting this up for others to look at.


coderanger:
Included in #54

PR #51, opened 2019-02-11
Proposal: Implement CR logger refactoring

blaggacao:


coderanger:
🤩 Was planning on working on this this week anyway. I think this is a great start! The next steps would require the long-awaited API changes to give components a more formal "name" so that we could improve the dev log formatter (and also to automatically include the component name for prod) since right now a lot of that is super manual. Also what do you think about `ctx.Log` to keep it a bit shorter as a commonly referenced value?

blaggacao:
Cool, I'm really happy to hear all that! `ctx.Log` looks better, indeed :smile:

blaggacao:
If I understand the idea of `errors.Wrap` right, in most situations catching errors at the `kubebilder.controller` level (with appropriate stack trace) looks to me like an idea worth enforcing and leave the internal logger to the info logs.

Maybe it is possible to inject the component name into that logger, somehow?

coderanger:
@blaggacao I am no longer involved with Ridecell however this has been implemented in my new coderanger/controller-utils library. Sorry for the delay in responding.

PR #53, opened 2019-02-12
Added command to migrations template for database flavor imports

Dhouti:


pkg/apis/summon/v1beta1/summonplatform_types.go
```diff
@@ -94,6 +94,12 @@ type SummonPlatformSpec struct {
 	// Database-related settings.
 	// +optional
 	Database DatabaseSpec `json:"database,omitempty"`
+	// Bucket that SummonFlavor data is stored in
+	// +optional
+	FlavorBucket string `json:"flavorBucket,omitempty"`
```
coderanger: Make this a const in the component since we don't have a reason to make it configurable.

pkg/apis/summon/v1beta1/summonplatform_types.go
```diff
@@ -94,6 +94,12 @@ type SummonPlatformSpec struct {
...
+	// The flavor of summon data to be imported upon creation
+	// +optional
+	SummonFlavor string `json:"summonFlavor,omitempty"`
```
coderanger: s/summon// because everything here is about summon

pkg/controller/summon/components/migrations.go
```diff
@@ -71,7 +77,25 @@ func (comp *migrationComponent) Reconcile(ctx *components.ComponentContext) (com
 		return components.Result{StatusModifier: setStatus(summonv1beta1.StatusDeploying)}, nil
 	}
 
-	obj, err := ctx.GetTemplate(comp.templatePath, nil)
+	var urlStr string
+	if instance.Spec.SummonFlavor != "" {
+		svc := s3.New(session.Must(session.NewSession()))
+		req, _ := svc.GetObjectRequest(&s3.GetObjectInput{
+			Bucket: aws.String(instance.Spec.FlavorBucket),
+			Key:    aws.String(fmt.Sprintf("%s.json", instance.Spec.SummonFlavor)),
```
coderanger: +.bz2 because compressed

pkg/controller/summon/templates/migrations.yml.tpl
```diff
@@ -31,7 +31,11 @@ spec:
         command:
         - sh
         - "-c"
+        {{- if ne .Extra.presignedUrl "" }}
+        - python manage.py migrate && python manage.py loadflavor {{ .Extra.presignedUrl }}
```
coderanger: Might want to make this `'{{ .Extra.presignedUrl }}'`, in case of weird chars in the URL string.

coderanger: Also I think this is going need the `--if-empty` or whatever you ended up calling it on the Django side so it doesn't fail if the DB has stuff in it :)

PR #59, opened 2019-03-14
AWS Integration tests

Dhouti:
Integration tests for iamuser and s3bucket


.circleci/config.yml
```diff
@@ -11,11 +11,12 @@ jobs:
         keys:
         - v1-vendor-{{ checksum "Gopkg.lock" }}
         - v1-vendor
-
+    - run: head -c 512 /dev/urandom | tr -dc 'a-z0-9' | head -c 8 > rand_owner_prefix
     - run: sudo mkdir /usr/local/kubebuilder/
     - run: curl -L https://storage.googleapis.com/kubebuilder-release/kubebuilder_master_linux_amd64.tar.gz | sudo tar -zxv --strip-components=1 -C /usr/local/kubebuilder/
     - run: make dep test
     - run: goveralls -coverprofile=gover.coverprofile -service=circle-ci -repotoken=$COVERALLS_TOKEN
+    - run: go run scripts/aws_cleanup/delete_operator_resources.go
```
coderanger: Needs a `when` so it runs even if the tests fail.

.circleci/config.yml
```diff
@@ -11,11 +11,12 @@ jobs:
         keys:
         - v1-vendor-{{ checksum "Gopkg.lock" }}
         - v1-vendor
-
+    - run: head -c 512 /dev/urandom | tr -dc 'a-z0-9' | head -c 8 > rand_owner_prefix
```
coderanger: So we can still use an env var to expose this to tests, the reason to put it in a file is each `run` is in its own shell so an `export` in one won't affect the next IIRC.

pkg/controller/s3bucket/controller_test.go
```diff
@@ -17,23 +17,239 @@ limitations under the License.
...
+			 }]
+		}`, bucketName)
+		c.Create(s3Bucket)
+
+		Eventually(func() error { return bucketExists() }, timeout).Should(Succeed())
+		Eventually(func() error { return bucketHasValidTag() }, timeout).Should(Succeed())
+		Eventually(func() string { return getBucketPolicy() }, timeout).Should(MatchJSON(s3Bucket.Spec.BucketPolicy))
+
+		fetchBucket := &awsv1beta1.S3Bucket{}
+		c.EventuallyGet(helpers.Name("test"), fetchBucket, c.EventuallyStatus(awsv1beta1.StatusReady))
```
coderanger: Move this first, and then you shouldn't need to `Eventually` the other checks since when it says Ready, things should be all done :)

.circleci/config.yml
Dhouti: Will be added in next PR that includes finalizers.

coderanger: :+1:

PR #63, opened 2019-04-01
Devops 1545 rabbitmq users and vhost

mehtapratikd:
Just renamed the user-api branch


pkg/controller/rabbitmquser/controller.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	dbv1beta1 "github.com/Ridecell/ridecell-operator/pkg/apis/db/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	"github.com/Ridecell/ridecell-operator/pkg/components"
+	rmqucomponents "github.com/Ridecell/ridecell-operator/pkg/controller/rabbitmquser/components"
+)
+
+// Add creates a new DjangoUser Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
```
Dhouti: "djangouser"

coderanger:
Basic functionality of making vhosts and users is in place. Still needs user permissions filled in so that the created user can actually do stuff and vhost policy support to set up HA.

pkg/controller/rabbitmquser/components/secret.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+type secretComponent struct{}
+
+func NewSecret() *secretComponent {
+	return &secretComponent{}
+}
+
+func (_ *secretComponent) WatchTypes() []runtime.Object {
+	return []runtime.Object{}
```
coderanger: Note to self: this should have Secret in it.

PR #66, opened 2019-04-17
Rds controller

Dhouti:


pkg/controller/rds/components/rds_instance.go
```diff
@@ -0,0 +1,345 @@
...
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() != rds.ErrCodeDBInstanceNotFoundFault {
+			return components.Result{}, errors.Wrapf(err, "rds: unable to describe db instance")
+		}
+		databaseNotExist = true
+	}
+
+	if databaseNotExist {
+		password = createDBPassword()
```
coderanger: How does this get persisted the first time?

pkg/controller/rds/components/security_group.go
```diff
@@ -0,0 +1,203 @@
...
+		},
+	})
+	if err != nil {
+		return components.Result{}, errors.Wrap(err, "rds: failed to describe security group")
+	}
+
+	if len(describeSecurityGroupsOutput.SecurityGroups) < 1 {
+		_, err = comp.ec2API.CreateSecurityGroup(&ec2.CreateSecurityGroupInput{
+			GroupName:   aws.String(instance.Name),
+			Description: aws.String("Created by ridecell-operator"),
```
coderanger: Maybe sprintf the instance Name and Namespace into this to make it easier to trace where things came from?

pkg/controller/rds/controller_test.go
```diff
@@ -0,0 +1,261 @@
...
+		c.Get(helpers.Name("test-rds.rds.credentials"), fetchSecret)
+		Expect(string(fetchSecret.Data["username"])).To(Equal("test_rds"))
+		Expect(len(string(fetchSecret.Data["password"]))).To(BeNumerically(">", 0))
+		Expect(len(string(fetchSecret.Data["endpoint"]))).To(BeNumerically(">", 0))
+
+		testContext := components.NewTestContext(fetchSecret, nil)
+
+		db, err := postgres.Open(testContext, &fetchRDS.Status.Connection)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(runTestQuery(db)).ToNot(HaveOccurred())
```
coderanger: Even though it's equivalent, I still usually pull things apart into `err = runTestQuery(db)` and `Expect(err).ToNot(HaveOccurred())` because it's visually easier to parse "error did not occur" than "query did not occur, oh wait no".

PR #67, opened 2019-04-22
Add policies for rabbitmq vhost

mehtapratikd:


pkg/test_helpers/fake_rabbitmq/fake_rabbitmq.go
```diff
@@ -65,3 +66,18 @@ func (frc *FakeRabbitClient) PutVhost(vhost string, _settings rabbithole.VhostSe
 	frc.Vhosts = append(frc.Vhosts, rabbithole.VhostInfo{Name: vhost})
 	return &http.Response{StatusCode: 201}, nil
 }
+
+func (frc *FakeRabbitClient) ListPoliciesIn(vhost string) (rec []rabbithole.Policy, err error) {
+	return frc.Policies, nil
+}
+
+func (frc *FakeRabbitClient) PutPolicy(vhost string, name string, policy rabbithole.Policy) (res *http.Response, err error) {
+	for key, policie := range frc.Policies {
```
coderanger: Because English is weird, this is `policy`.

pkg/controller/rabbitmq_vhost/components/vhost.go
```diff
@@ -81,13 +81,36 @@ func (comp *vhostComponent) Reconcile(ctx *components.ComponentContext) (compone
...
+	// 1. Get policies and compare patterns, if found update, else create.
+	// 2. Remove patterns that are not found
+
+	policiesList, err := rmqc.ListPoliciesIn(instance.Spec.VhostName)
+	if err != nil {
+		return components.Result{}, errors.Wrapf(err, "error fetching policies for vhost %s", instance.Spec.VhostName)
+	}
+	for policyName := range instance.Spec.Policies {
+		var pFound bool
+		for i := 0; i < len(policiesList); i++ {
```
coderanger: Any reason to not use a normal `for ... range` loop here?

pkg/controller/rabbitmq_vhost/components/vhost.go
```diff
@@ -81,13 +81,36 @@ func (comp *vhostComponent) Reconcile(ctx *components.ComponentContext) (compone
...
+			if instance.Spec.Policies[policyName].Pattern == policiesList[i].Pattern {
```
coderanger: Will need to check for more than just Pattern, I would probably try using `reflect.DeepEqual` for simplicity.

pkg/test_helpers/fake_rabbitmq/fake_rabbitmq.go
mehtapratikd: yeah but that name already is used, part of the arguments

pkg/controller/rabbitmq_vhost/components/vhost.go
mehtapratikd: Nope, agree range is cleaner..

pkg/controller/rabbitmq_vhost/components/vhost.go
mehtapratikd: I think i am just going to check the policy name here.. nothing else...

pkg/apis/db/v1beta1/rabbitmqvhost_types.go
```diff
@@ -30,7 +30,7 @@ type RabbitmqPolicy struct {
 	Priority int `json:"priority,omitempty"`
 	// Additional arguments added to the entities (queues,
 	// exchanges or both) that match a policy
-	Definition string `json:"definition"`
+	Definition string `yaml:"definition"`
```
coderanger: Should be JSON, that's how K8s works.

pkg/controller/rabbitmq_vhost/components/vhost.go
```diff
@@ -81,13 +81,56 @@ func (comp *vhostComponent) Reconcile(ctx *components.ComponentContext) (compone
...
+			policiesList = policiesList[:len(policiesList)-1]
+		}
+		// Create/Update policy
+		policy := rabbithole.Policy{}
+		policy.Pattern = instance.Spec.Policies[policyName].Pattern
+		policy.ApplyTo = instance.Spec.Policies[policyName].ApplyTo
+		policy.Priority = instance.Spec.Policies[policyName].Priority
+		err := yaml.Unmarshal([]byte(instance.Spec.Policies[policyName].Definition), &policy.Definition)
+		if err != nil {
+			return components.Result{}, errors.Wrapf(err, "error unable to parse policy definition for %s", instance.Spec.VhostName)
```
coderanger: Rather than logging the VhostName, the important thing to include in the error is the policyName.

pkg/controller/rabbitmq_vhost/components/vhost.go
```diff
@@ -81,13 +81,56 @@ func (comp *vhostComponent) Reconcile(ctx *components.ComponentContext) (compone
 		}
 	}
 
+	// Policies
+	policiesList, err := rmqc.ListPoliciesIn(instance.Spec.VhostName)
+	if err != nil {
+		return components.Result{}, errors.Wrapf(err, "error fetching policies for vhost %s", instance.Spec.VhostName)
+	}
+	for policyName := range instance.Spec.Policies {
```
coderanger: I think you

pkg/controller/rabbitmq_vhost/components/vhost.go
```diff
@@ -81,13 +81,56 @@ func (comp *vhostComponent) Reconcile(ctx *components.ComponentContext) (compone
...
+		var pFound bool
+		var index int
+		actualPolicyName := fmt.Sprintf("%s-%s", instance.Spec.VhostName, policyName)
+		for i := range policiesList {
```
coderanger: I think you forgot about the two-argument form of `range`:

```go
for policyName, policy := range instance.Spec.Policies {
```

pkg/controller/rabbitmq_vhost/components/vhost.go
```diff
@@ -81,13 +81,56 @@ func (comp *vhostComponent) Reconcile(ctx *components.ComponentContext) (compone
...
+		if err != nil {
+			return components.Result{}, errors.Wrapf(err, "error unable to parse policy definition for %s", instance.Spec.VhostName)
+		}
+		_, err = rmqc.PutPolicy(instance.Spec.VhostName, actualPolicyName, policy)
+		if err != nil {
+			return components.Result{}, errors.Wrapf(err, "error updating policy for vhost %s", instance.Spec.VhostName)
+		}
+	}
+	// Remove policies for a vhost which are not in the Spec
+	for i := range policiesList {
```
coderanger: Similar to the above, we don't care about the index, just the name, `for _, policyName := range ...`

pkg/controller/rabbitmq_vhost/components/vhost.go
```diff
@@ -81,13 +81,56 @@ func (comp *vhostComponent) Reconcile(ctx *components.ComponentContext) (compone
...
+		var index int
+		actualPolicyName := fmt.Sprintf("%s-%s", instance.Spec.VhostName, policyName)
+		for i := range policiesList {
+			if actualPolicyName == policiesList[i].Name {
+				pFound = true
+				index = i
+				break
+			}
+		}
+		if pFound {
```
coderanger: Generally better to put this check in the delete loop to keep things more visible.

pkg/controller/summon/templates/rabbitmq/vhost.yml.tpl
```diff
@@ -3,3 +3,22 @@ kind: RabbitmqVhost
...
+  policies:
+    HA:
+      pattern: ^(?!amq\.).*
+      apply-to: queues
+      priority: -10
+      definition: |
+        ha-mode: exactly
+        ha-params: 2
+        ha-sync-mode:	automatic
+    Federate:
```
coderanger: This isn't needed, the federation policies were for the cool-eel migration.

pkg/controller/rabbitmquser/components/user_test.go
```diff
@@ -35,6 +36,24 @@ import (
 var _ = Describe("RabbitmqUser User Component", func() {
 	comp := rmqucomponents.NewUser()
 	var frc *fake_rabbitmq.FakeRabbitClient
+	spec1 := dbv1beta1.RabbitmqUserSpec{
```
coderanger: Doing a static-y initialization of a mutable object is generally unwise in tests, because it could get changed in one test leading to incorrect results in the next. This is why you usually do the setup in a Before so that it is re-init'd for each test.

pkg/controller/rabbitmq_vhost/templates/rabbitmq_user.yml.tpl
```diff
@@ -4,4 +4,10 @@ metadata:
   name: {{ .Instance.Name }}
   namespace: {{ .Instance.Namespace }}
 spec:
-  username: {{ .Instance.Spec.VhostName}}-user
+  username: {{ .Instance.Spec.VhostName }}-user
+  tags: policymaker
```
coderanger: It shouldn't actually need this tag AFAIK.

pkg/controller/rabbitmq_vhost/components/vhost.go
mehtapratikd: umm It's easier to remove the policy from the list when it's already being compared ..

PR #68, opened 2019-04-22
remove region hardcoding in migrations component

Dhouti:


Dhouti:
This is hardcoded for a reason, all flavors live in a bucket in us-west-2.

PR #69, opened 2019-04-24
Added PostgresUser Type and Controller

Dhouti:


pkg/apis/db/v1beta1/postgresuser_types.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"github.com/Ridecell/ridecell-operator/pkg/apis/helpers"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// PostgresUserSpec defines the desired state of PostgresUser
+type PostgresUserSpec struct {
+	Connection PostgresConnection `json:"database"`
+	Username   string             `json:"username"`
```
coderanger: Can be optional with default

pkg/apis/db/v1beta1/postgresuser_types.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// PostgresUserStatus defines the observed state of PostgresUser
+type PostgresUserStatus struct {
+	Status            string            `json:"status"`
+	Message           string            `json:"message"`
+	PasswordSecretRef helpers.SecretRef `json:"passwordSecretRef"`
```
coderanger: Use "Connection PostgresConnection `json:"connection"`" instead and fill in the hostname and port from the Spec.Connection.

pkg/controller/postgresuser/components/postgresuser.go
```diff
@@ -0,0 +1,140 @@
...
+
+type PostgresUserComponent struct {
+}
+
+func NewPostgresUser() *PostgresUserComponent {
+	return &PostgresUserComponent{}
+}
+
+func (_ *PostgresUserComponent) WatchTypes() []runtime.Object {
+	return []runtime.Object{}
```
coderanger: Should watch the secret since it does depend on it.

pkg/controller/postgresuser/components/postgresuser.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (_ *PostgresUserComponent) IsReconcilable(ctx *components.ComponentContext) bool {
+	instance := ctx.Top.(*dbv1beta1.PostgresUser)
+
+	if instance.Status.SecretStatus != dbv1beta1.StatusReady {
```
coderanger: Not really needed, components run in order.

pkg/controller/postgresuser/components/postgresuser.go
```diff
@@ -0,0 +1,140 @@
...
+		// 28P01 == invalid password
+		if pqerr, ok := err.(*pq.Error); ok && pqerr.Code == "28P01" {
+			invalidPassword = true
+		} else {
+			return components.Result{}, errors.Wrap(err, "postgres_user: failed to query database")
+		}
+	}
+
+	if invalidPassword {
+		query := fmt.Sprintf(`ALTER USER %s WITH PASSWORD %s;`, instance.Spec.Username, fetchSecret.Data[instance.Status.PasswordSecretRef.Key])
```
coderanger: Potential bobby tables. Use `?` or `$1` if possible, if user has to be an identifier rather than a string than might need to use `QuoteIdentifier`.

PR #77, opened 2019-05-13
- Add liveness probe for celeryd

mehtapratikd:


pkg/controller/summon/templates/celeryd/deployment.yml.tpl
```diff
@@ -1,6 +1,63 @@
...
+        imagePullPolicy: Always
+        command: [python, "-m", celery, "-A", summon_platform, worker, "-l", info]
+        ports:
+        - containerPort: 8000
+        resources:
+          requests:
+            memory: 512M
+            cpu: 500m
+          limits:
+            memory: 1G
```
coderanger: Should be 3G as of this morning.

mehtapratikd: 👍

PR #84, opened 2019-05-21
Iamuser access key convergance

Dhouti:


pkg/controller/iamuser/components/iam_user.go
```diff
@@ -227,17 +227,24 @@ func (comp *iamUserComponent) Reconcile(ctx *components.ComponentContext) (compo
...
-			return components.Result{}, errors.Wrapf(err, "iam_user: failed to list access keys")
-		}
-		// Delete access keys if they exist
-		for _, accessKeyMeta := range existingAccessKeys.AccessKeyMetadata {
+		fetchAccessKeyID = []byte{}
+	}
+
+	var foundAccessKeyID bool
+	for _, accessKeyMeta := range existingAccessKeys.AccessKeyMetadata {
+		if aws.StringValue(accessKeyMeta.AccessKeyId) == string(fetchAccessKeyID) && len(fetchAccessKeyID) != 0 {
```
coderanger: Checking both of these (string and len) seems redundant?

PR #86, opened 2019-05-23
Ci refactor, again

coderanger:


coderanger:
I'm back on my CI bullshit. This makes the first job only run unit tests so it takes a bit less time (1-2m) and runs the control plane via CircleCI rather than envtest. This is a minor speedup for a few cases, and lets us run controller-manager too, so we have GC which might help reduce OOM issues too.

PR #87, opened 2019-05-29
Devops 1667 vanity aliases

rc-marina:
1) Added `Aliases` to SummonPlatformSpec.
2) Modified `ingress.yml.tpl` to create IngressRule for each individual alias using the same service as the primary hostname.
3) Tested new changes in `ingress_type.go` and `controller_test.go`

### DIFF
--- a/pkg/controller/summon/controller_test.go
+++ b/pkg/controller/summon/controller_test.go
@@ -343,44 +343,6 @@ var _ = Describe("Summon controller", func() {
 		}, timeout).Should(Succeed())
 	})
 
-	It("reconciles ingress service with specified vanity hostnames", func() {
-		c := helpers.TestClient
-		instance := &summonv1beta1.SummonPlatform{ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: helpers.Namespace},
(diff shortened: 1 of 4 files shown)


pkg/controller/summon/controller_test.go
```diff
@@ -343,6 +343,44 @@ var _ = Describe("Summon controller", func() {
...
+		c.EventuallyGet(helpers.Name("foo-web"), service)
+		Expect(service.Spec.Ports[0].Port).To(BeEquivalentTo(8000))
+
+		// Check the web Ingress object.
+		ingress := &extv1beta1.Ingress{}
+		c.EventuallyGet(helpers.Name("foo-web"), ingress)
+
+		Expect(ingress.Spec.TLS[0].SecretName).To(Equal("foo-tls"))
+
+		for _, vanityName := range instance.Spec.Aliases {
```
rc-marina: This does the same validation as in `ingress_test.go`, but inside controller test to ensure the SummonPlatformSpec results in the creation of a correct Ingress object. 

Is this okay, or sort of an unnecessary duplicate of ingress test?

PR #88, opened 2019-06-02
Switch to an ObjectRef so we can more easily do cross-database DbConfig references

coderanger:
This will be needed for microservices.


jengo:
I don't know Go that well, so I am not the best for review ... if you want to get someone else to review it go for it.  Otherwise ya go for it.

PR #90, opened 2019-06-05
Rds snapshot controller

Dhouti:


.circleci/config.yml
```diff
@@ -132,13 +132,28 @@ jobs:
     - *kube-apiserver
     - *kube-controller-manager
     environment:
-      GINKGO_ARGS: -focus @rds
+      GINKGO_ARGS: -focus @rds -skip @rdssnapshot
```
coderanger: Rather than getting into weird regex battles, maybe we can just make the tag be `@snapshot`? That won't conflict with anything else as a prefix :)

pkg/apis/db/v1beta1/rdssnapshot_types.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"time"
+)
+
+// RDSSnapshotSpec defines the desired state of RDSSnapshot
+type RDSSnapshotSpec struct {
+	RDSInstanceID string `json:"rdsInstanceID"`
+	//+kubebuilder:validation:Pattern=^[a-zA-Z][a-zA-Z0-9-]*[a-zA-Z0-9]$
```
coderanger: 💯

pkg/apis/db/v1beta1/rdssnapshot_types.go
```diff
@@ -0,0 +1,66 @@
...
+	// +optional
+	SnapshotID string `json:"SnapshotID,omitempty"`
+	// +optional
+	TTL time.Duration `json:"ttl,omitempty"`
```
coderanger: Heh, this wouldn't have worked before but it should now with the upgrade to controller-tools 0.1.10?

pkg/apis/db/v1beta1/rdssnapshot_types.go
```diff
@@ -0,0 +1,66 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"time"
+)
+
+// RDSSnapshotSpec defines the desired state of RDSSnapshot
+type RDSSnapshotSpec struct {
+	RDSInstanceID string `json:"rdsInstanceID"`
+	//+kubebuilder:validation:Pattern=^[a-zA-Z][a-zA-Z0-9-]*[a-zA-Z0-9]$
+	// +optional
+	SnapshotID string `json:"SnapshotID,omitempty"`
```
coderanger: Convention is for JSON field names to start with a lowercase.

pkg/controller/rdssnapshot/components/defaults.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (_ *defaultsComponent) IsReconcilable(_ *components.ComponentContext) bool {
+	return true
+}
+
+func (comp *defaultsComponent) Reconcile(ctx *components.ComponentContext) (components.Result, error) {
+	instance := ctx.Top.(*dbv1beta1.RDSSnapshot)
+
+	if instance.Spec.SnapshotID == "" {
+		creationTimestamp := instance.ObjectMeta.CreationTimestamp.Add(time.Second * 0)
```
coderanger: The `Add(0)` is weird enough that it probably warrants a comment.

pkg/controller/rdssnapshot/components/defaults_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	rdssnapshotcomponents "github.com/Ridecell/ridecell-operator/pkg/controller/rdssnapshot/components"
+	. "github.com/Ridecell/ridecell-operator/pkg/test_helpers/matchers"
+)
+
+var _ = Describe("rds Defaults Component", func() {
```
coderanger: These need actual tests.

pkg/controller/rdssnapshot/components/rds_snapshot.go
```diff
@@ -0,0 +1,193 @@
...
+	return true
+}
+
+func (comp *RDSSnapshotComponent) Reconcile(ctx *components.ComponentContext) (components.Result, error) {
+	instance := ctx.Top.(*dbv1beta1.RDSSnapshot)
+
+	if instance.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !helpers.ContainsFinalizer(RDSSnapshotFinalizer, instance) {
+			instance.ObjectMeta.Finalizers = helpers.AppendFinalizer(RDSSnapshotFinalizer, instance)
+			err := ctx.Update(ctx.Context, instance.DeepCopy())
```
coderanger: Why the DeepCopy?

pkg/controller/rdssnapshot/components/rds_snapshot.go
```diff
@@ -0,0 +1,193 @@
...
+		},
+		&rds.Tag{
+			Key:   aws.String("scheduled-for-deletion"),
+			Value: aws.String(fmt.Sprintf("%v", scheduledDelete)),
+		},
+	}
+	if scheduledDelete {
+		deletionTimestampTag := &rds.Tag{
+			Key:   aws.String("deletion-timestamp"),
+			Value: aws.String(deletionTimestamp),
```
coderanger: This string is only set is the TTL is not 0, otherwise will be "". Is that okay?

pkg/controller/rdssnapshot/components/rds_snapshot.go
```diff
@@ -0,0 +1,193 @@
...
+	return []runtime.Object{}
+}
+
+func (_ *RDSSnapshotComponent) WatchChannel() chan event.GenericEvent {
+	genericChannel := make(chan event.GenericEvent)
+	return genericChannel
+}
+
+func (_ *RDSSnapshotComponent) IsReconcilable(ctx *components.ComponentContext) bool {
+	return true
```
coderanger: What happens if we try to initiate a snapshot on a database that isn't ready yet? Will it correctly error or does something go boom?

pkg/controller/rdssnapshot/controller.go
```diff
@@ -0,0 +1,77 @@
...
+
+	err = c.Controller.Watch(
+		&source.Channel{Source: genericChannel},
+		&handler.EnqueueRequestForObject{},
+	)
+	return err
+}
+
+func watchTTL(watchChannel chan event.GenericEvent, k8sClient client.Client) {
+	rdsSnapshots := &dbv1beta1.RDSSnapshotList{}
```
coderanger: I think maybe you meant to have this whole thing in a `for { ... }` loop? This will just run once I think?

pkg/controller/rdssnapshot/controller_suite_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/Ridecell/ridecell-operator/pkg/controller/rdssnapshot"
+	"github.com/Ridecell/ridecell-operator/pkg/test_helpers"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/rds"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestTemplates(t *testing.T) {
```
coderanger: Not templates :)

pkg/controller/rdssnapshot/components/rds_snapshot.go
Dhouti: This block also only triggers if the TTL is not 0. The plan was to only apply this tag if the snapshot has a TTL, otherwise the tag name doesn't have much value. Could always switch to setting it to something like "false" if there is no deletion planned.

pkg/controller/rdssnapshot/controller_suite_test.go
Dhouti: ?

pkg/controller/rdssnapshot/components/rds_snapshot.go
Dhouti: CreateDBSnapshot will throw `ErrCodeInvalidDBInstanceStateFault`

pkg/apis/db/v1beta1/rdssnapshot_types.go
```diff
@@ -0,0 +1,66 @@
...
+	SnapshotID string `json:"snapshotID,omitempty"`
+	// +optional
+	TTL time.Duration `json:"ttl,omitempty"`
+}
+
+// RDSSnapshotStatus defines the observed state of RDSSnapshot
+type RDSSnapshotStatus struct {
+	Status     string `json:"status"`
+	Message    string `json:"message"`
+	SnapshotID string `json:"instanceID"`
```
coderanger: I think you meant to say snapshotId for the field name?

pkg/controller/rdssnapshot/controller_suite_test.go
coderanger: This isn't testing Templates. the actual name of the function doesn't matter to Go as long as it starts with `Test`, and we've gotten a bit copy-pasta happy with it :)

coderanger:
:shipit:

PR #91, opened 2019-06-06
Devops 1668 deployment status integration

rc-marina:
Post to deployment status tool only when SummonPlatform object successfully created. (Does not cover updating the `was_successful` flag.)


pkg/controller/summon/controller_notification_test.go
```diff
@@ -204,104 +195,275 @@ var _ = Describe("Summon controller", func() {
...
+			Expect(deployStatusServer.ReceivedRequests()).Should(HaveLen(2))
+		})
+
+		It("does not send post request for deploy when something went wrong", func() {
+			c := helpers.TestClient
+
+			// Change up tag expectation.
+			instance.Spec.Version = "1.2.3"
+			// Set up verification handler to check our request body.
+			deployStatusServer.AppendHandlers(
```
coderanger: Strictly speaking this isn't needed since the point of the test is that no requests are made :)

rc-marina: Hah, I agree. Will cut out the handlers /extra code lines and leave it to `Expect(deployStatusServer.ReceivedRequests()).Should(HaveLen(0))`

rc-marina:
Previous test_other failed with:

`    sends a single post request on deploy, even with subsequent reconciles [It]
    /go/src/github.com/Ridecell/ridecell-operator/pkg/controller/summon/controller_notification_test.go:370

    JSON Mismatch
    Expected
        <string>: {
          "customer_name": "statustester",
          "deploy_user": "ridecell-operator",
          "environment": "test-62c5e4e88d081010f226",
          "tag": "1-abcdef1-master"
        }
    to match JSON of
        <string>: {
          "customer_name": "notifytest2",
          "deploy_user": "ridecell-operator",
          "environment": "test-1bfe21ae6db3e800f9c3",
          "tag": "80813-eb6b515-master"
        }
    
    first mismatched key: "customer_name"`

But now its 
`Summon controller
/go/src/github.com/Ridecell/ridecell-operator/pkg/controller/summon/controller_test.go:44
  manages the status correctly [It]
  /go/src/github.com/Ridecell/ridecell-operator/pkg/controller/summon/controller_test.go:346

  Timed out after 30.000s.
  Expected
      <string>: Error
  to equal
      <string>: Ready`

Flakey test we'll probably have to investigate and fix >_<
In mean time, pushing forward because GTR!

PR #99, opened 2019-06-13
Use first value in the aliases array as WEB_URL

mehtapratikd:


pkg/controller/summon/components/defaults.go
```diff
@@ -136,7 +136,11 @@ func (comp *defaultsComponent) Reconcile(ctx *components.ComponentContext) (comp
 	defVal("CACHE_URL", "redis://%s-redis/1", instance.Name)
 	defVal("FIREBASE_ROOT_NODE", "%s", instance.Name)
 	defVal("TENANT_ID", "%s", instance.Name)
-	defVal("WEB_URL", "https://%s", instance.Spec.Hostname)
+	if instance.Spec.Aliases != nil {
```
coderanger:
```suggestion
        webUrl := instance.Spec.Hostname
	if instance.Spec.Aliases != nil && len(instance.Spec.Aliases) > 0 {
	  webUrl = instance.Spec.Aliases[0]
        }
        defVal("WEB_URL", "https://%s", webUrl)
```

coderanger: Or something like that? The important bit is the len check so it doesn't crash on `aliases: []` :)

PR #100, opened 2019-06-18
Devops 1719 multi slack channels

rc-marina:


pkg/controller/summon/components/notification.go
```diff
@@ -185,6 +185,17 @@ func (c *notificationComponent) handleSuccess(instance *summonv1beta1.SummonPlat
 		}
 	}
 
+	// Send to additional slack channels.
+	if len(instance.Spec.Notifications.SlackChannels) != 0 {
```
coderanger: Do you need to check the len first? Usually range on an empty slice should be safe I think.

pkg/controller/summon/controller_notification_test.go
```diff
@@ -69,6 +74,21 @@ var _ = Describe("Summon controller", func() {
 				},
 			},
 		}
+
+		// Set up Slack client with the test user credentials and find the most recent message.
+		slackClient = slack.New(os.Getenv("SLACK_API_KEY"))
```
coderanger: You took out the Skip stuff, which is still needed or anyone running these tests on their workstation without the API key will get failures.

pkg/controller/summon/components/notification.go
rc-marina: you were right -- we dont need the if check

pkg/controller/summon/controller_notification_test.go
rc-marina: Added skip back in with following commit c05363f

rc-marina:
After discussing with @coderanger about the tradeoffs of saving time on running Slack and Deployment-status tests (by combining their Contexts) vs having cleaner code if we were to separate them using Contexts, it didn't seem worthwhile to complicate the code unless it shaved off more than a min of test run time. 

For comparison:
If Slack and Deployment-status test are split into context, my local test times (focused on controller_notification_test) are:

```
Ran 8 of 17 Specs in 84.076 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 9 Skipped
PASS | FOCUSED

Ginkgo ran 1 suite in 1m31.169810504s
Test Suite Passed
```
If combined:
```Ran 4 of 13 Specs in 48.055 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 9 Skipped
PASS | FOCUSED

Ginkgo ran 1 suite in 53.594165424s
```

PR #102, opened 2019-06-20
Add Deployment for prometheus exporter 

mehtapratikd:


pkg/controller/shared_components/postgres/postgres.go
```diff
@@ -223,6 +223,24 @@ func (comp *postgresComponent) reconcileLocal(ctx *components.ComponentContext,
 	return res, existing.Status.String(), conn, nil
 }
 
+func (comp *postgresComponent) reconcileExporter(ctx *components.ComponentContext, conn *dbv1beta1.PostgresConnection) (components.Result, error) {
+	var existing *appsv1.Deployment
```
coderanger: We don't actually need these outside the inner function so you can just `:=` them down there and save a few lines :)

pkg/controller/shared_components/postgres/postgres.go
```diff
@@ -136,7 +136,7 @@ func (comp *postgresComponent) Reconcile(ctx *components.ComponentContext) (comp
 	if err != nil {
 		return res, err
 	}
-
+	comp.reconcileExporter(ctx, conn)
```
coderanger: Need to check the error response from this.

pkg/controller/shared_components/postgres/templates/postgres-exporter.yml.tpl
```diff
@@ -0,0 +1,36 @@
...
+      labels:
+        app: postgres-exporter
+        instance: {{ .Instance.Name }}
+    spec:
+      containers:
+      - name: postgres-exporter
+        image: wrouesnel/postgres_exporter:v0.4.7
+        env:
+        - name: DATA_SOURCE_URI
+          value: "{{ .Extras.Conn.Host }}:{{ .Extras.Conn.Port }}/?sslmode=disable"
```
coderanger: There is an `SSLMode` field in the connection struct you can use to know if you should disable things.

PR #104, opened 2019-06-21
Allow overriding the RDS instance ID for a database

coderanger:
Mostly matters for restoring from backups but maybe for migrating existing summon prod instances where the ID doesn't match our standard format.


coderanger:
Thinking further, I don't actually like this approach. This really belongs on the SummonPlatform object as the main data store. Going to try that out.

coderanger:
Replaced by #141

PR #112, opened 2019-07-09
Webui v1

Dhouti:


pkg/webui/actions/auth.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	membershipCheckURL := fmt.Sprintf("https://api.github.com/orgs/Ridecell/members/%s", user.NickName)
+	req, err := http.NewRequest("GET", membershipCheckURL, nil)
+	if err != nil {
+		return err
+	}
+	req.Header.Add("Authorization", fmt.Sprintf("token %s", user.AccessToken))
+
+	req1, err := http.NewRequest("GET", "https://api.github.com/repos/Ridecell/kubernetes-summon", nil)
```
coderanger: What does this get used for? Looks like it's just printed out so I'm guessing it's from testing?

pkg/webui/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,136 @@
...
+	err := apis.AddToScheme(scheme.Scheme)
+	if err != nil {
+		panic(err)
+	}
+}
+
+var k8sclient *client.Client
+
+func getClient() (client.Client, error) {
+	if k8sclient != nil {
```
coderanger: This could result in race conditions. Though the losing side of the race would just have done a bunch of extra work.

pkg/webui/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	mapper, err := apiutil.NewDiscoveryRESTMapper(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	newClient, err := client.New(cfg, client.Options{Scheme: scheme.Scheme, Mapper: mapper})
```
coderanger: This just gets us a plain client, which is good for now, but I think long term we're going to want to use the cached informer-driven client layer on top of this. But it will require some more explicit initialization stuffs.

PR #116, opened 2019-07-16
Quick version of an init container for RabbitMQ stuff for microservices.

coderanger:


coderanger:
Test failure is a known flake.

PR #120, opened 2019-07-18
Add RidecellIngress type and controller

vivekjainx86:


pkg/apis/addtoscheme_ingress_v1beta1.go
```diff
@@ -0,0 +1,26 @@
+/*
+Copyright 2018-2019 Ridecell, Inc.
```
coderanger: It isn't 2018 :)

pkg/apis/ingress/v1beta1/ridecellingress_types.go
```diff
@@ -0,0 +1,52 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
+	v1beta1 "k8s.io/api/extensions/v1beta1"
```
coderanger: It's customary in Kubernetes code to use a name like `extv1beta1` when importing API packages like this, so it's clearer in the code where things came from.

pkg/apis/ingress/v1beta1/ridecellingress_types_test.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"testing"
+
+	"github.com/onsi/gomega"
+	"golang.org/x/net/context"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestStorageRidecellIngress(t *testing.T) {
```
coderanger: We use Ginkgo for our tests rather than the default GoTest, can check out some of the other packages for examples :)

pkg/controller/ridecellingress/ridecellingress_controller.go
```diff
@@ -0,0 +1,213 @@
+/*
```
coderanger: This will have to rewritten using our systems, we do controllers a bit differently from plain Kubebuilder :) @mehtapratikd can walk you through it.

pkg/controller/ridecellingress/ridecellingress_controller.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func populateDomainName(hostname string, instance *ingressv1beta1.RidecellIngress) string {
+	domain := "ridecell.io"
+	//Check wheather hostname is a full domain name or not
+	if !strings.Contains(hostname, ".") {
+		//Populate the cluster domain name according to cloud, region and environment
```
coderanger: This isn't quite the logic we want. If the hostname field contains a `.`, then we don't want to modify it all. If it doesn't, then we want to convert it to `$hostname.$clustername.ridecell.io`. There are two ways to get the cluster name. The simple one is if the `ridecell.io/cluster-name` label is set, use that. Otherwise the cluster name is `region-environment`.

pkg/controller/ridecellingress/ridecellingress_controller_suite_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var cfg *rest.Config
+
+func TestMain(m *testing.M) {
```
coderanger: Similar to the type tests, we have some more complex testing tools we use :)

pkg/controller/ridecellingress/ridecellingress_controller.go
coderanger: Ignore me, your algorithm is better. So the only change would be to not change hostnames with `.`s because they are probably already full names.

pkg/apis/addtoscheme_ingress_v1beta1.go
```diff
@@ -0,0 +1,26 @@
+/*
+Copyright 2019-2020 Ridecell, Inc.
```
coderanger: It isn't 2020, yet.

pkg/controller/ridecellingress/components/defaults.go
```diff
@@ -0,0 +1,43 @@
+/*
```
coderanger: This component is not needed anymore, it doesn't do anything.

pkg/controller/ridecellingress/components/ridecellingress_test.go
```diff
@@ -0,0 +1,95 @@
...
+						},
+						SecretName: "custom-tls",
+					},
+				},
+			},
+		}
+		ctx = components.NewTestContext(&instance, ridecellingress.Templates)
+	})
+
+	It("creates an RidecellIngress object using above template", func() {
```
coderanger: Could use some more unit tests here.

pkg/controller/ridecellingress/controller_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	AfterEach(func() {
+		if CurrentGinkgoTestDescription().Failed {
+			helpers.DebugList(&ingressv1beta1.RidecellIngressList{})
+			helpers.DebugList(&extv1beta1.Ingress{})
+		}
+		helpers.TeardownTest()
+	})
+
+	It("Creates RidecellIngress kind", func() {
```
coderanger: I would convert most of these controller tests to component unit tests to save time. Controller tests are far slower.

pkg/controller/ridecellingress/ridecellingress_controller.go
```diff
@@ -0,0 +1,213 @@
+/*
+Copyright 2018-2019 Ridecell, Inc.
```
coderanger: This whole file can be removed now.

pkg/controller/ridecellingress/ridecellingress_controller_suite_test.go
coderanger: File can be removed.

pkg/controller/ridecellingress/templates/ingress.yml.tpl
```diff
@@ -0,0 +1,11 @@
+{
```
coderanger: I would probably make this slightly more normal YAML like the other templates.

PR #122, opened 2019-07-18
Add ability to halt deployments after migrations for x time

Dhouti:


Dhouti:
There's also a fix in here to prevent redis restarts on version change

pkg/controller/summon/components/deployment.go
```diff
@@ -58,7 +58,7 @@ func (comp *deploymentComponent) IsReconcilable(ctx *components.ComponentContext
 	if instance.Status.PostgresStatus != dbv1beta1.StatusReady {
 		return false
 	}
-	if instance.Status.MigrateVersion != instance.Spec.Version {
+	if instance.Status.Status != summonv1beta1.StatusDeploying {
```
coderanger: It should also process if in status ready.

pkg/controller/summon/components/statefulset.go
```diff
@@ -55,7 +55,7 @@ func (comp *statefulsetComponent) IsReconcilable(ctx *components.ComponentContex
 	if instance.Status.PostgresStatus != dbv1beta1.StatusReady {
 		return false
 	}
-	if instance.Status.MigrateVersion != instance.Spec.Version {
+	if instance.Status.Status != summonv1beta1.StatusDeploying {
```
coderanger: Ready here too.

pkg/controller/summon/components/deployment.go
Dhouti: Add StatusReady check to IsReconcilable() && Move StatusDeploying check to beginning of Reconcile()?

PR #123, opened 2019-07-19
[WIP] Big upgrade from controller-runtime 0.1 to 0.2.

coderanger:
Most of the changes fall into two categories: CreateOrUpdate's MutateFn no longer takes the existing object as a parameter (instead it's a mutated closure) and client.List now has the options using the same style as Delete, with the options at the end and built by factory functions.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -773,20 +773,24 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:8ce639e43d430e23970c8b6c3a0f90e880654566f1a70be28066e354e51139b1"
+  digest = "1:e1879ae997f4d391287eafadc0f1140e004d2212fc788ddf10ffab7d48bfcbc4"
   name = "golang.org/x/tools"
   packages = [
     "go/ast/astutil",
     "go/gcexportdata",
-    "go/internal/cgo",
     "go/internal/gcimporter",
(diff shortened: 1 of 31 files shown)


Gopkg.lock
```diff
@@ -1058,47 +1132,52 @@
...
+    "pkg/loader",
+    "pkg/markers",
     "pkg/rbac",
-    "pkg/util",
     "pkg/webhook",
   ]
   pruneopts = "T"
-  revision = "b8adde9bc6d7f3fba449d306613a9daed23676c8"
-  version = "v0.1.10"
+  revision = "7e14e77d06eab03bd24f518142f422cace857708"
```
rc-marina: got a complaint  during make dep:
`Failed to write sigs.k8s.io/controller-tools (from https://github.com/coderanger/controller-tools.git)@embedded-error
grouped write of manifest, lock and vendor: error while writing out vendor tree: failed to write dep tree: failed to export sigs.k8s.io/controller-tools: fatal: failed to unpack tree object 7e14e77d06eab03bd24f518142f422cace857708`

Not sure how Gopkg.toml and Gopkg.lock stuff work, but I tried using the latest revision for embedded-error branch, `b09fb5439a8f7ee536631e3d79a2ec1a6a31e05d` and the complaint goes away
¯\_(ツ)_/¯

PR #129, opened 2019-08-01
Cross namespace postgresdatabase fix

Dhouti:


jengo:
I am currently blocked by this issue

pkg/controller/postgresdatabase/components/secret.go
```diff
@@ -0,0 +1,87 @@
...
+		newSecret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fetchSecret.Name,
+				Namespace: instance.Namespace,
+			},
+			Type: corev1.SecretTypeOpaque,
+			Data: fetchSecret.Data,
+		}
+
+		_, err = controllerutil.CreateOrUpdate(ctx.Context, ctx, newSecret, func(existingObj runtime.Object) error {
```
rc-marina: Is this creating a copy of the secret from DbConfig's namespace into the PostgresDB's namespace so whenever there's a `PasswordSecretRef.Resolve` call, it'll be able to find the copy?

Re-paste of PasswordSecretRef's Resolve():

```
func (s *SecretRef) Resolve(ctx *components.ComponentContext, defaultKey string) (string, error) {
	namespace := ctx.Top.(metav1.Object).GetNamespace()
	key := s.Key
	if key == "" {
		key = defaultKey
	}

	fetch := &corev1.Secret{}
	err := ctx.Get(ctx.Context, types.NamespacedName{Name: s.Name, Namespace: namespace}, fetch)
	if err != nil {
		return "", errors.Wrapf(err, "unable to fetch secret %s/%s", namespace, s.Name)
	}

	val, ok := fetch.Data[key]
	if !ok {
		return "", errors.Errorf("key %s not found in secret %s/%s", key, namespace, s.Name)
	}

	return string(val), nil
}
```

rc-marina: Could we modify Resolve() above to take an optional `namespace` argument, and if that's not set, default to checking the current namespace?

I think that would save us from having to create a copy of the secret in another namespace? But I'm not sure if this modification alone is enough to completely fix/resolve the issue we're encountering though.

Dhouti: That is the goal, yes. :)

The optional Namespace argument was the first proposed approach, Noah and I opted not to go that route after some chatting.

PR #130, opened 2019-08-03
Add sumologic links to webui

Dhouti:


coderanger:
Secondary table due to wonky alignment issues, can be fixed in CSS later.

PR #137, opened 2019-08-15
Add postgres to initcontainer and make all operations optional

Dhouti:


cmd/initcontainer/main.go
```diff
@@ -135,19 +189,39 @@ func Update(env, serviceName string, c client.Client, data map[string]interface{
...
+	if !k8serr.IsNotFound(err) {
+		pgdbConnection := pgdb.Status.Connection
 
+		// Fetch the postgres database password
+		pgdbPassword, err := pgdbConnection.PasswordSecretRef.Resolve(ctx, "password")
+		if err != nil {
+			return err
+		}
+
+		data["DATABASE"].(map[interface{}]interface{})["PASSWORD"] = pgdbPassword
```
coderanger: This should cope with the `DATABASE` key itself not existing too.

cmd/initcontainer/main.go
```diff
@@ -121,7 +144,38 @@ func DumpYAML(data map[string]interface{}) error {
...
+	pgdb := &dbv1beta1.PostgresDatabase{}
+	err := ctx.Get(ctx.Context, types.NamespacedName{Namespace: serviceName, Name: fmt.Sprintf("svc-%s-%s", env, serviceName)}, pgdb)
+	if err != nil && !k8serr.IsNotFound(err) {
+		return err
+	}
+
+	if !k8serr.IsNotFound(err) {
+		pgdbConnection := pgdb.Status.Connection
+
+		dbField := data["DATABASE"].(map[interface{}]interface{})
```
coderanger: As below, this will segfault if the key isn't found.

Dhouti:
Also added IAMuser to initcontainer

PR #142, opened 2019-08-22
Handle race condition

Dhouti:


pkg/controller/summon/components/backup.go
```diff
@@ -88,35 +86,34 @@ func (comp *backupComponent) Reconcile(ctx *components.ComponentContext) (compon
...
+		return components.Result{StatusModifier: func(obj runtime.Object) error {
+			instance := obj.(*summonv1beta1.SummonPlatform)
+			instance.Status.Status = summonv1beta1.StatusMigrating
+			instance.Status.BackupVersion = instance.Spec.Version
+			return nil
+		}}, nil
+	}
+
+	// Unknown status, reqeueue
+	return components.Result{Requeue: true}, nil
```
coderanger: Probably use a short delay, otherwise this will check every few milliseconds.

PR #151, opened 2019-09-10
Revamp replicas settings and also fix up lots of summon component tes…

coderanger:
…ts to be more like real usage.


coderanger:
New config would look like:

```
replicas:
  web: 8
  celeryd: 12
```

PR #152, opened 2019-09-10
Try out a static analysis checker.

coderanger:
Will probably fail hilariously on our current code.


coderanger:
Can't review because github rules but looks good to me

PR #154, opened 2019-09-10
Trying to fix dep issues

coderanger:
Now that CircleCI caching is back on, it's causing weird dep issues. Trying to fix those.


coderanger:
This works as evidenced by the passing build, but I don't really understand why or how. We should probably just accept it an move on, and try to switch to Modules asap.

PR #155, opened 2019-09-10
updated new-relic config

vivekjainx86:
Update summon new-relic configuration with reference to https://github.com/Ridecell/devops/blob/433f16e4f154da14dd2e265d95f5dcac9356bde0/ansible/roles/summon-platform-newrelic/tasks/main.yml#L13

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,65 +2,9 @@
 
 
 [[projects]]
-  branch = "default"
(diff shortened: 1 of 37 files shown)


coderanger:
We're trying to get away from using this file. While we can merge this as a stop-gap, we should also start a similar config file in summon-platform to be the authoritative source.

avdhoot:
also needed for CORE-1620

PR #161, opened 2019-09-17
Sqs region stuff

coderanger:
This should further streamline SQS configuration I think.

Only major remaining question is if qa instances should default to the qa queue.

### DIFF
--- a/pkg/controller/summon/components/defaults.go
+++ b/pkg/controller/summon/components/defaults.go
@@ -93,7 +93,7 @@ func (comp *defaultsComponent) Reconcile(ctx *components.ComponentContext) (comp
 		switch instance.Spec.Environment {
 		case "prod", "uat":
 			switch instance.Spec.AwsRegion {
-			case "eu-entral-1":
+			case "eu-central-1":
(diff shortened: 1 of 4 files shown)


coderanger:
Fixes DEVOPS-1982 and DEVOPS-1984

pkg/controller/summon/components/defaults.go
```diff
@@ -83,18 +83,29 @@ func (comp *defaultsComponent) Reconcile(ctx *components.ComponentContext) (comp
...
+		}
 	}
 	if instance.Spec.SQSQueue == "" {
-		if instance.Spec.Environment == "prod" || instance.Spec.Environment == "uat" {
-			instance.Spec.SQSQueue = "prod-data-pipeline"
-		} else {
+		switch instance.Spec.Environment {
+		case "prod", "uat":
+			switch instance.Spec.AwsRegion {
+			case "eu-entral-1":
```
rc-marina: Ack, typo -- we're missing the 'c'

coderanger: Good catch!

PR #162, opened 2019-09-17
check secret in namespace for Exclusive mode

vivekjainx86:


pkg/controller/postgresdatabase/components/secret.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+http://www.apache.org/licenses/LICENSE-2.0
```
coderanger: Intentional?

pkg/controller/postgresdatabase/components/secret.go
```diff
@@ -48,34 +48,53 @@ func (_ *secretComponent) Reconcile(ctx *components.ComponentContext) (component
...
-		fetchSecret := &corev1.Secret{}
-		err := ctx.Client.Get(ctx.Context, types.NamespacedName{
-			Name:      instance.Status.AdminConnection.PasswordSecretRef.Name,
-			Namespace: instance.Spec.DbConfigRef.Namespace,
-		}, fetchSecret)
+		dbconfig := &dbv1beta1.DbConfig{}
+		err := ctx.Get(ctx.Context, types.NamespacedName{Name: instance.Spec.DbConfigRef.Name, Namespace: instance.Spec.DbConfigRef.Namespace}, dbconfig)
 		if err != nil {
-			return components.Result{}, errors.Wrap(err, "secret: unable to get dbconfig secret")
+			return components.Result{}, errors.Wrapf(err, "secret: unable to get dbconfig %s/%s for PostgresDatabase %s", instance.Spec.DbConfigRef.Namespace, instance.Spec.DbConfigRef.Name, instance.Name)
```
coderanger: So we don't generally put stuff like instance.Name in the error message since it's inherently tied to the object already, and the name of the secret will generally be in the inner error but that's more debatable.

pkg/controller/postgresdatabase/components/secret.go
```diff
@@ -48,34 +48,53 @@ func (_ *secretComponent) Reconcile(ctx *components.ComponentContext) (component
 
 	// If namespace is not defaulted and does not match no action is needed
 	if instance.Spec.DbConfigRef.Namespace != "" && instance.Spec.DbConfigRef.Namespace != instance.Namespace {
-		fetchSecret := &corev1.Secret{}
-		err := ctx.Client.Get(ctx.Context, types.NamespacedName{
-			Name:      instance.Status.AdminConnection.PasswordSecretRef.Name,
-			Namespace: instance.Spec.DbConfigRef.Namespace,
-		}, fetchSecret)
+		dbconfig := &dbv1beta1.DbConfig{}
+		err := ctx.Get(ctx.Context, types.NamespacedName{Name: instance.Spec.DbConfigRef.Name, Namespace: instance.Spec.DbConfigRef.Namespace}, dbconfig)
```
coderanger: This needs to use the same defaulting as https://github.com/Ridecell/ridecell-operator/blob/master/pkg/controller/shared_components/postgres/postgres.go#L353-L363, which maybe means we should make that function public.

pkg/controller/postgresdatabase/components/secret.go
```diff
@@ -48,34 +48,53 @@ func (_ *secretComponent) Reconcile(ctx *components.ComponentContext) (component
...
 		if err != nil {
-			return components.Result{}, errors.Wrap(err, "secret: unable to get dbconfig secret")
+			return components.Result{}, errors.Wrapf(err, "secret: unable to get dbconfig %s/%s for PostgresDatabase %s", instance.Spec.DbConfigRef.Namespace, instance.Spec.DbConfigRef.Name, instance.Name)
 		}
 
-		newSecret := &corev1.Secret{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      fetchSecret.Name,
+		if dbconfig.Spec.Postgres.Mode == "Exclusive" {
+			// the secret must in the current namespace only
```
coderanger: No copying is needed in exclusive mode, you can just bail out entirely.

pkg/controller/postgresdatabase/components/secret.go
vivekjainx86: I didn't get this point, here I am getting instance of DBConfig to get 'Mode' value, the function you mentioned is just returning dbconfig name and its namespace.

pkg/controller/postgresdatabase/components/secret.go
vivekjainx86: In this block, I am not copying secret, just confirming that secret is present in current namespace.

pkg/controller/postgresdatabase/components/secret.go
vivekjainx86: not intentional, its due to auto indent

pkg/controller/postgresdatabase/components/secret.go
vivekjainx86: got it

pkg/controller/postgresdatabase/components/secret.go
coderanger: You don't need to confirm that, it doesn't accomplish anything.

pkg/controller/postgresdatabase/components/secret.go
coderanger: Yes, but the ref name and namespace can be `""` in which case you need to use default values.

coderanger: Actually never mind, since the default namespace case means this doesn't need to activate, so you can just check that directly. Just need to check it.

pkg/controller/postgresdatabase/components/secret.go
vivekjainx86: ok, removing this block

PR #167, opened 2019-09-25
Add support for the new web runtime goop

coderanger:
Requires https://github.com/Ridecell/summon-platform/pull/10500). With backwards compat for older versions because that __main__.py file is new.


coderanger:
This is dumb, why did I do this.

PR #171, opened 2019-10-04
Enable self service monitoring for summon

avdhoot:
Will update this PR regelulry. This will help me to include suggestion provided by team in early phase.


pkg/apis/summon/v1beta1/summonplatform_types.go
```diff
@@ -277,3 +281,8 @@ type SummonPlatformList struct {
 func init() {
 	SchemeBuilder.Register(&SummonPlatform{}, &SummonPlatformList{})
 }
+
+// MonitorSpec will enable in monitoring. (In future we can use it to configure monitor.ridecell.io)
+type MonitorSpec struct {
+	Enabled bool `json:"enabled,omitempty"`
```
avdhoot: Absens of slack channel/channels monitoring will not enabled even value of `enabled` is `true` in

PR #175, opened 2019-10-11
New kind for mock car server tenant

vivekjainx86:


avdhoot:
Feedback provided offline.

pkg/apis/summon/v1beta1/mockcarservertenant_types.go
```diff
@@ -0,0 +1,59 @@
...
+package v1beta1
+
+import (
+	//corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type MockCarServerTenantSpec struct {
+	// Important: Run "make" to regenerate code after modifying this file
+	TenantHardwareType string `json:"tenanthardwaretype,omitempty"`
```
coderanger: Generally the JSON fields using camel case, like `tenantHardwardType`.

pkg/apis/summon/v1beta1/summonplatform_types.go
```diff
@@ -188,6 +188,14 @@ type SummonPlatformSpec struct {
 	// Enable monitoring
 	// +optional
 	Monitoring MonitorSpec `json:"monitoring,omitempty"`
+	// Enable mock car server
+	// +optional
+	EnableMockCarServer bool `json:"enableMockCarServer,omitempty"`
+	// If Mock car server enabled, provide tenant hardware type
+	// +optional
+	// +kubebuilder:validation:Enum=OTAKEYS;MENSA
+	MockTenantHardwareType string `json:"mockTenantHardwareType,omitempty"`
```
coderanger: Is there a reason to use two fields for this instead of just one? If the hardware type is set, we can assume that means they want to enable the mock server? The main value is you set the hardware type to a default so it streamlines the simple case but I'm worried that might come back to bite us later. Not a big deal either way.

pkg/controller/mockcarservertenant/components/mockcarservertenant.go
```diff
@@ -0,0 +1,125 @@
...
+	// Get data from secret
+	postData := map[string]string{}
+	postData["api_key"] = string(otakeysSecret.Data["OTAKEYS_API_KEY"])
+	postData["sec_key"] = string(otakeysSecret.Data["OTAKEYS_SECRET_KEY"])
+	postData["api_token"] = string(otakeysSecret.Data["OTAKEYS_TOKEN"])
+	postData["push_api_key"] = string(otakeysSecret.Data["OTAKEYS_PUSH_API_KEY"])
+	postData["push_sec_key"] = string(otakeysSecret.Data["OTAKEYS_PUSH_SECRET_KEY"])
+	postData["push_token"] = string(otakeysSecret.Data["OTAKEYS_PUSH_TOKEN"])
+
+	// Build callback url
```
coderanger: This should get provided in since the hostname on a SummonPlatform can be overridden.

pkg/controller/mockcarservertenant/components/secret.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	res.StatusModifier = func(obj runtime.Object) error {
+		instance := obj.(*summonv1beta1.MockCarServerTenant)
+		instance.Status.KeysSecretRef = secretName
+		return nil
+	}
+	return res, err
+}
+
+func RandStringBytes(n int) []byte {
```
coderanger: There is a shared copy of something like this under utils.RandomString

pkg/controller/summon/components/app_secrets.go
```diff
@@ -184,6 +185,18 @@ func (comp *appSecretComponent) Reconcile(ctx *components.ComponentContext) (com
 			appSecretsData[k] = string(v)
 		}
 	}
+	// If OTAKEYS_API_KEY is not present, add ota keys generated by mockcarservertenant kind
+	apiKey, ok := appSecretsData["OTAKEYS_API_KEY"]
+	if (!ok || len(apiKey.(string)) == 0) && instance.Spec.EnableMockCarServer {
```
coderanger: If the mock server is enabled explicitly, I feel like it should always be used. Or maybe if both are specified, it should be an error?

pkg/controller/summon/templates/mockcarservertenant.yml.tpl
```diff
@@ -0,0 +1,7 @@
+apiVersion: summon.ridecell.io/v1beta1
+kind: MockCarServerTenant
+metadata:
+  name: {{ .Instance.Name }}
+  namespace: {{ .Instance.Namespace }}
+spec:
+  TenantHardwareType: {{ .Instance.Spec.MockTenantHardwareType }}
```
coderanger: This doesn't match the capitalization you used in the `json` tags in the struct.

pkg/controller/mockcarservertenant/components/mockcarservertenant.go
vivekjainx86: yes, I am doing this change :)

pkg/controller/mockcarservertenant/components/secret.go
vivekjainx86: there were no digits in utils.RandomString character set, thats why I written this. But I confirmed from Hardware team that digits are optional, using this implementation.

vivekjainx86: I tried to use utils.RandomByte implementation but it returned 44 bytes of data instead of 32. Something might be wrong with it, I kept this logic.

coderanger: Heh, the number of bytes requested is the number of actual random bytes, which then get encoded into a safe character set which abuses some base64 encoding code :)

PR #178, opened 2019-10-15
Fix app secrets not updating when input secrets change

Dhouti:


Dhouti:
Defaulting for `instance.Spec.Secrets` has been moved inside of the app_secrets component.

PR #182, opened 2019-10-16
Flag for enabling prometheus exporter behavior for web pods.

Dhouti:


avdhoot:
i thik we should use monitring flag insted creating this flag for every componet. As sout goal is make monitoring standerd. @coderanger

coderanger:
@avdhoot This is a temporary thing while we're testing the system because the updated exporter code in summon isn't available everywhere yet. Once everywhere is upgraded past PCR 14, we can remove the flag :)

avdhoot:
@coderanger  I think we can shift  that logic in the app. try to run/check `summon_platform` if failed run `summon_platform.wsgi.application`

coderanger:
@avdhoot I would rather not because if anything goes wrong, we would have no way to turn it off. Feature flags are a good thing :)

avdhoot:
@coderanger we do not have feature flag on every standerd feature. I do not see there will bug in if condition :)

PR #183, opened 2019-10-16
Add servicemonitor component

Dhouti:


pkg/controller/summon/components/servicemonitor.go
```diff
@@ -0,0 +1,62 @@
+/*
+Copyright 2018-2019 Ridecell, Inc.
```
coderanger: It's not 2018

pkg/controller/summon/components/servicemonitor.go
```diff
@@ -0,0 +1,62 @@
...
+func (comp *serviceMonitorComponent) IsReconcilable(ctx *components.ComponentContext) bool {
+	return true
+}
+
+func (comp *serviceMonitorComponent) Reconcile(ctx *components.ComponentContext) (components.Result, error) {
+	instance := ctx.Top.(*summonv1beta1.SummonPlatform)
+
+	// If our flag is not set or is false exit early
+	if instance.Spec.Metrics.Web == nil || !*instance.Spec.Metrics.Web {
+		return components.Result{}, nil
```
coderanger: If you want to be extra fancy, this could delete things too.

PR #186, opened 2019-10-18
Implement anti-affinity for Summon pods

rc-marina:


rc-marina:
`podAntiAffinity.podAffinityTerm.topologyKey: failure-domain.beta.kubernetes.io/zone` means pod cannot be scheduled onto a node if that node is in the same zone as a pod with the label `app.kubernetes.io/instance: {{ .Instance.Name }}-{{ block "componentName" . }}{{ end }}`

PR #189, opened 2019-10-19
Delete pod disruption budget before reconcile

Dhouti:


Dhouti:
Not 100% happy with this solution, it's a temporary hack. It was either this (or similar more verbose code), a deletion before every reconcile, or a custom createorupdate function.

PR #194, opened 2019-10-23
add mock tenant secret to inputSecret list

vivekjainx86:


coderanger:
This would work but I think we can do better than this

vivekjainx86:
> This would work but I think we can do better than this



> This would work but I think we can do better than this

The problem here is, in fetchSecret(), if a secret not found, it returns error, and in our case, tenant secret's existence is conditional (enableMockCarServer flag), and if thats secret is not present, then it will throw error

PR #198, opened 2019-10-24
Added more monitoring

avdhoot:


pkg/controller/summon/templates/monitoring.yml.tpl
```diff
@@ -33,4 +33,36 @@ spec:
...
+    - alert: Pods are not running
+      expr: kube_pod_container_status_running{namespace={{ .Instance.Namespace | quote }}, pod=~"{{ .Instance.Name }}.*" ,pod!~ "{{ .Instance.Name }}-migrations-.*"} 
+      for: 3m
+      labels:
+        severity: critical
+        servicename: {{ .Instance.Name }}
+      annotations:
+        summary: "{{ $labels.pod }} pod is not running."
+    - alert: Memory Critical
+      expr: container_memory_rss{namespace={{ .Instance.Namespace | quote }}, pod=~"{{ .Instance.Name }}-.*" }  / on(pod, container) kube_pod_container_resource_limits_memory_bytes{namespace={{ .Instance.Namespace | quote }}, pod=~"{{ .Instance.Name }}-.*"}  * 100 > 80
```
coderanger: Any reason to use RSS instead of `container_memory_usage_bytes`?

PR #200, opened 2019-10-30
reconcile database before periscope

rc-marina:


coderanger:
Fixes a very niche bug where we changed the database config for an existing PostgresDatabase object, so it carried through the existing status rather than resetting everything to `""`. This resulted in an ordering problem.

PR #202, opened 2019-11-05
add support for skipping finalizers

vivekjainx86:
Skipping finalizer's execution by using _ridecell.io/skip-finalizer: "true"_ annotation to following types:
IAM User
MockCarServerTenant
Monitor
RDSInstance
RDS Snapshot
S3 bucket

### DIFF
--- a/pkg/controller/iamuser/components/iam_user_test.go
+++ b/pkg/controller/iamuser/components/iam_user_test.go
@@ -18,6 +18,7 @@ package components_test
 
 import (
 	"context"
+	"os"
 	"time"
 
 	. "github.com/Ridecell/ridecell-operator/pkg/test_helpers/matchers"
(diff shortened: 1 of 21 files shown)


pkg/controller/iamuser/components/iam_user_test.go
```diff
@@ -152,6 +153,8 @@ var _ = Describe("iam_user aws Component", func() {
 	})
 
 	Describe("finalizer tests", func() {
+		os.Setenv("ENABLE_FINALIZERS", "true")
```
coderanger: You probably want to do this in the `It` block (or in a `BeforeEach`) so that it runs before the test. Putting it here will run it once before the whole process starts :) Ginkgo ordering is a bit weird.

vivekjainx86: I want it to set env var before whole process starts :) , thats why I put that line directly inside describe{}

coderanger: No, that will not work if, for example, we added another test that sets the value to false. You should fix this.

PR #212, opened 2019-11-14
Remove --fast from golangci-lint

Dhouti:


coderanger:
https://github.com/golangci/golangci-lint/pull/758 has successfully cut our RAM usage to about a third of the previous value, which is low enough to run all linters on CI!

PR #222, opened 2019-12-19
Fix djangouser creation

Dhouti:
A new NOT NULL column was added to the common_staff table causing new django user creations to fail. 
https://github.com/Ridecell/summon-platform/pull/10486


Dhouti:
Needs some additional planning: Will break on any summon database that is not at least on PCR15

Dhouti:
The likelihood of new deployments using the versions that would cause issues is low enough that we're comfortable moving forward.

PR #227, opened 2019-12-31
Autodeploy is almost here!

rc-marina:
Review of logic and maybe additional tests that come to mind? Then we can run it in sandbox environment and see how it goes! :O

### DIFF
--- a/pkg/apis/summon/v1beta1/summonplatform_types.go
+++ b/pkg/apis/summon/v1beta1/summonplatform_types.go
@@ -140,8 +140,12 @@ type SummonPlatformSpec struct {
 	// Hostname aliases (for vanity purposes)
 	// +optional
 	Aliases []string `json:"aliases,omitempty"`
-	// Summon image version to deploy.
+	// Summon image version to deploy. If this isn't specified, AutoDeploy must be.
+	// +optional
 	Version string `json:"version"`
+	// Branch to watch for new images and auto-deploy.
+	// +optional
+	AutoDeploy string `json:"autoDeploy,omitempty"`
 	// Name of the secret to use for secret values.
 	Secrets []string `json:"secrets,omitempty"`
 	// Name of the secret to use for image pulls. Defaults to `"pull-secret"`.
--- a/pkg/apis/summon/v1beta1/summonplatform_types.go
+++ b/pkg/apis/summon/v1beta1/summonplatform_types.go
@@ -142,7 +142,7 @@ type SummonPlatformSpec struct {
 	Aliases []string `json:"aliases,omitempty"`
 	// Summon image version to deploy. If this isn't specified, AutoDeploy must be.
 	// +optional
-	Version string `json:"version"`
+	Version string `json:"version,omitempty"`
 	// Branch to watch for new images and auto-deploy.
 	// +optional
 	AutoDeploy string `json:"autoDeploy,omitempty"`
--- a/pkg/controller/summon/components/autodeploy.go
+++ b/pkg/controller/summon/components/autodeploy.go
@@ -17,26 +17,26 @@ limitations under the License.
 package components
 
 import (
-	"github.com/Ridecell/ridecell-operator/pkg/components"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 
 	summonv1beta1 "github.com/Ridecell/ridecell-operator/pkg/apis/summon/v1beta1"
+	"github.com/Ridecell/ridecell-operator/pkg/components"
 	gcr "github.com/Ridecell/ridecell-operator/pkg/utils/gcr"
 )
 
 type AutoDeployComponent struct {
-	tagFetcher func()
+	tagFetcher func(string) (string, error)
 }
 
 func NewAutoDeploy() *AutoDeployComponent {
 	return &AutoDeployComponent{
-		tagFetcher: gcr.GetSummonTags,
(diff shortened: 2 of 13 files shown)


pkg/apis/summon/v1beta1/summonplatform_types.go
```diff
@@ -140,8 +140,12 @@ type SummonPlatformSpec struct {
 	// Hostname aliases (for vanity purposes)
 	// +optional
 	Aliases []string `json:"aliases,omitempty"`
-	// Summon image version to deploy.
+	// Summon image version to deploy. If this isn't specified, AutoDeploy must be.
+	// +optional
 	Version string `json:"version"`
```
coderanger: For it to be optional, you have to add `,omitempty`.

pkg/controller/summon/components/autodeploy.go
```diff
@@ -0,0 +1,86 @@
...
+func (comp *AutoDeployComponent) Reconcile(ctx *components.ComponentContext) (components.Result, error) {
+	instance := ctx.Top.(*summonv1beta1.SummonPlatform)
+	branchRegex, err := gcr.SanitizeBranchName(instance.Spec.AutoDeploy)
+
+	if err != nil {
+		return components.Result{}, errors.Wrapf(err, "Failed to sanitize AutoDeploy: %s for docker image search", instance.Spec.AutoDeploy)
+	}
+
+	// Fetch tags from gcr. This triggers cache check and possibly updates tags before assigning version for deployment.
+	comp.tagFetcher()
```
coderanger: What do you think about merging this with GetLatestImageOfBranch? Or rather, have dependency injection point be on GetLatestImageOfBranch so we can inject a mock version of that for tests rather than having some explicit global state-y things? The real version of it would then call the cache update internally.

pkg/controller/summon/components/deployment.go
```diff
@@ -76,6 +76,15 @@ func (comp *deploymentComponent) Reconcile(ctx *components.ComponentContext) (co
 		return components.Result{}, nil
 	}
 
+	// Set error status to prevent further deployments until it is resolved.
+	if instance.Spec.Version == "" && instance.Spec.AutoDeploy == "" {
```
coderanger: We should probably do this check earlier, like in the defaults component, since the version is used for other stuff before we get this far (backup names, etc). Also don't need to use a status modifier, can just return `errors.Errorf("Spec.Version OR Spec.AutoDeploy must be set. No Version set for deployment.")` or similar.

pkg/controller/summon/controller.go
```diff
@@ -109,5 +123,44 @@ func Add(mgr manager.Manager) error {
...
+			for _, summonInstance := range summonInstances.Items {
+				if summonInstance.Spec.AutoDeploy == "" {
+					continue
+				}
+				watchChannel <- event.GenericEvent{Object: &summonInstance, Meta: &summonInstance}
+			}
+			// We checked all summonInstances for autodeploy. Update lastChecked.
+			lastChecked = time.Now()
+		}
+		time.Sleep(time.Millisecond * 100)
```
coderanger: Can probably bump this to more like 10 seconds to reduce the number of calls to GCR.

pkg/utils/gcr/gcr.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func GetLatestImageOfBranch(branchTag string) (string, error) {
+	var latestImage string
+	latestBuild := 0
+	for _, image := range CachedTags {
+
+		// Append $ so we do not match beyond the end of branchTag. This prevents
+		// situations where we have similar branch names like "fix-for-ticket1" and "fix-for-ticket2"
+		match, err := regexp.Match(branchTag+"$", []byte(image))
```
coderanger: Use `regexp.QuoteMeta(branchTag)` just in case someone manages to get a regexp metachar in there somehow. Shouldn't be possible since we remove all of them beforehand, but easy to be double safe :D

pkg/controller/summon/components/autodeploy.go
rc-marina: Instead of our tests actually depending on the actual gcr globals var states( `LastCacheUpdate`, `CacheTags` and `CacheExpiry`), we would have Mock version of these only visible to the test scope?

That probably makes it easier for me to manipulate testing cache timing conditions!
Will experiment and push the revised version if all goes well.

coderanger: The unit test for this controller wouldn't even have to mock those specifically, it could just return static list and call it a day. The unit tests for the gcr package would need to make sure those bits of code work though.

pkg/controller/summon/components/deployment.go
rc-marina: Will SummonPlatformStatus `Status` and `Message` be updated without the Status Modifier, or we're just leaving it out?

coderanger: Yep, just like for all errors :) Specifically it calls SetError in the status interface which sets both of those.

rc-marina: Thanks! Added a test case that confirms we get the error in commit 994a2e1

pkg/controller/summon/components/autodeploy.go
rc-marina: Merged `GetSummonTags` logic with `GetLatestImageOfBranch` and revised testing around it in commit 0785dfb. Hopefully better now. Let me know what you think!

rc-marina:
Going to create an Issue for the flakey `triggers autodeploy reconcile when tag cache updated (via watcher)` test case.

PR #233, opened 2020-01-13
CORE-1813: Fixing python2 to python3 migration caching issues

rakesh-ridecell:
* Added new config's ASGI_URL_V2 and CACHE_URL_V2
* Updated test


coderanger:
Why is the existing fix not sufficient? At a minimum this is definitely not required or advisable for the Channels broker. That uses msgpack if it isn't stable across Python versions I'm going to have to have a very long talk with the author :) The rest of the cache should not be using pickled objects.

coderanger:
Discussed in Slack, probably better to use the built-in cache VERSION option in Django as that is at least tied more directly to Summon. Still not a great solution since we need to remember to manually bump that version again in the future when we do other things that break stored pickles, such as the modules branch for PCR 2.

PR #234, opened 2020-01-14
✨ Automatically delete the beat state file if it exits with a non-0 exit code

coderanger:
This should catch many kinds of corruption and is what we will always do manually first anyway.


coderanger:
I _think_ this should work smoothly with a normal pod restart (i.e. not delete) according to https://github.com/celery/celery/blob/master/celery/apps/beat.py#L151-L156. But we can check for the echo in the log in sandbox.

PR #235, opened 2020-01-14
🐛 Fix beat file path to delete.

coderanger:
I should check on these things before writing code.


Dhouti:
I should trust my gut more during reviews. :)

PR #241, opened 2020-01-23
🐛 Go pq doesn't like postgis://

coderanger:


coderanger:
The code in question: https://github.com/lib/pq/blob/d6156e141ac6c06345c7c73f450987a9ed4b751f/connector.go#L60

PR #251, opened 2020-02-05
automate reporting postgresuser

rc-marina:


rc-marina:
When this gets merged to sandbox, any microservices and summon instances hosted on aws-us-sandbox (dev, qa) that has an existing reporting user in the database will have the reporting user password replaced.

To keep things running normally, svc-reporting credentials will need to be updated with the reporting user pw that ridecell-operator generates.

The same will be true when this makes it into production environment.

Having `ridectl reporting` implementation (copy/paste `ridectl periscope`) would make the updates easier.

PR #252, opened 2020-02-06
Debug autodeploy delay

rc-marina:


rc-marina:
So I think I broke the caching behavior when I committed 0785dfb and missed a piece to the logic. 

The cache tag update behavior has to happen independently from `GetLatestImageOfBranch` since before this fix, the watch channel was depending on on the condition:
```if lastChecked.IsZero() || gcr.LastCacheUpdate.After(lastChecked)``` to trigger reconciles, 
but `gcr.LastCacheUpdate.After(lastChecked)` would never be true because `gcr.LastCacheUpdate` doesn't change until an AutoDeploy Reconcile is triggered.

AutoDeploy Reconcile doesn't happen unless the controller sent an event to trigger it, so we were in a circular dependency state.

The only other time AutoDeploy Reconcile would be trigger besides the channel event is if the SummonPlatform .yaml spec changed. 

I'm not quite sure yet what caused SummonPlatform .yaml spec to change (or maybe something else occuring) which results in autodeploy actually deploying ~1 hr after a docker Image became available.

rc-marina:
abandoning for #254. I git rebased and merged and the resulting git log feels messy.

PR #254, opened 2020-02-11
fix summonplatform controller awareness of autodeploy caching logic

rc-marina:
replaces #252, since I did a git rebase and merge in #252 and history looks funky to me.
This one is cleaner...


rc-marina:
😞still running into flakey failures due to timing issues. Will have a separate PR to nail it down, but for now, would like to push autodeploy timing fix out.

PR #262, opened 2020-03-05
Iamrole controller

Dhouti:


pkg/controller/iamrole/components/iam_role.go
```diff
@@ -0,0 +1,361 @@
...
+
+func (comp *iamRoleComponent) Reconcile(ctx *components.ComponentContext) (components.Result, error) {
+	instance := ctx.Top.(*awsv1beta1.IAMRole)
+
+	// do the template thing on all the stuff
+	templateData := templatingData{
+		Region: os.Getenv("AWS_REGION"),
+	}
+
+	roleName, err := templateData.parseField(instance.Spec.RoleName)
```
coderanger: Little weird to hang this off the template data struct rather than the component struct.

pkg/controller/iamrole/components/iam_role.go
```diff
@@ -0,0 +1,361 @@
...
+	var foundOperatorTag, foundKiamTag bool
+	for _, tags := range listRoleTagsOutput.Tags {
+		if aws.StringValue(tags.Key) == "ridecell-operator" {
+			foundOperatorTag = true
+		}
+		if aws.StringValue(tags.Key) == "Kiam" {
+			foundKiamTag = true
+		}
+	}
+	if !foundOperatorTag || !foundKiamTag {
```
coderanger: If !foundOperatorTag, the tagging will fail since we said this isn't allowed to adopt existing resources.

pkg/controller/iamrole/components/iam_role.go
```diff
@@ -0,0 +1,361 @@
...
+		})
+		if err != nil {
+			return components.Result{}, errors.Wrapf(err, "iam_role: unable to update assume role policy for %s", aws.StringValue(role.RoleName))
+		}
+	}
+
+	return components.Result{StatusModifier: func(obj runtime.Object) error {
+		instance := obj.(*awsv1beta1.IAMRole)
+		instance.Status.Status = awsv1beta1.StatusReady
+		instance.Status.Message = "Role exists"
```
coderanger: Would be good to sprintf the RoleName into this string since it's possibly dynamic. Just for human verification/sanity checking.

scripts/aws_cleanup/delete_operator_resources.go
```diff
@@ -246,6 +264,36 @@ func getIAMUsersToDelete(iamsvc *iam.IAM, prefix string) ([]*string, error) {
...
+func getIAMRolesToDelete(iamsvc *iam.IAM, prefix string) ([]*string, error) {
+	// List all the roles
+	listRolesOutput, err := iamsvc.ListRoles(&iam.ListRolesInput{})
+	if err != nil {
+		return nil, err
+	}
+
+	var iamRolesToDelete []*string
+	for _, role := range listRolesOutput.Roles {
+		regexString := fmt.Sprintf(`^%s-.*-summon-platform$`, prefix)
```
coderanger: Is this regex still accurate?

PR #264, opened 2020-03-10
First pass on GCPProject kind

Dhouti:


pkg/controller/gcpproject/components/defaults.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (_ *defaultsComponent) IsReconcilable(_ *components.ComponentContext) bool {
+	return true
+}
+
+func (comp *defaultsComponent) Reconcile(ctx *components.ComponentContext) (components.Result, error) {
+	instance := ctx.Top.(*gcpv1beta1.GCPProject)
+
+	// Fill in defaults.
+	if instance.Spec.EnableFirebase == nil {
```
coderanger: In the places where you use these, you already check for nil as equivalent to false so these are kind of redundant.

PR #268, opened 2020-04-01
[WIP] Implementing Vertical Pod Autoscaling

rc-marina:


rc-marina:
Starting all VPAs with `UpdateMode "Off"` (pod resources won't be changed), allowing us to see what is recommended by the VPA. 

`UpdateMode` is defined in each component rather than the helper template for flexibility, in the case we want to try different `UpdateMode` for different components, or change a few components at a time rather than all at once.

pkg/test_helpers/test_helpers.go
```diff
@@ -59,6 +66,51 @@ type PerTestHelpers struct {
...
+	var crds []*apiextv1beta1.CustomResourceDefinition
+	for _, doc := range docs {
+		crd := &apiextv1beta1.CustomResourceDefinition{}
+		if err = yaml.Unmarshal(doc, crd); err != nil {
+			return nil, err
+		}
+
+		// Seems we need to set ObjectMeta ourselves to workaround this error:
+		// CustomResourceDefinition.apiextensions.k8s.io "" is invalid: metadata.name: Required value: name or generateName is required
+		if crd.ObjectMeta.Name == "" {
```
Dhouti, line 105: leftover bits from old hacks

rc-marina: We actually still need this hack too or Controller tests still fails with that error :/

rc-marina: I was wrong. :) Was using the wrong yaml library for unmarshalling. Fixed and removed invalid hack!

pkg/controller/summon/components/vpa.go
```diff
@@ -0,0 +1,52 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package components
+
+import (
+	"github.com/Ridecell/ridecell-operator/pkg/components"
+	"k8s.io/apimachinery/pkg/runtime"
+	autoscaling "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/apis/autoscaling.k8s.io/v1"
```
coderanger, line 22: Should be `autoscalingv1` to match import convention.

pkg/controller/summon/controller.go
```diff
@@ -74,59 +74,70 @@ func Add(mgr manager.Manager) error {
 		summoncomponents.NewPVC("redis/volumeclaim.yml.tpl"),
 		summoncomponents.NewRedisDeployment("redis/deployment.yml.tpl"),
 		summoncomponents.NewService("redis/service.yml.tpl"),
+		summoncomponents.NewVPA("redis/vscaler.yml.tpl"),
```
coderanger, line 77: Please name the templates `vpa` to match the standard naming.

pkg/controller/summon/templates/helpers/vscaler.yml.tpl
```diff
@@ -0,0 +1,21 @@
...
+    app.kubernetes.io/component: {{ block "componentType" . }}{{ end }}
+    app.kubernetes.io/part-of: {{ .Instance.Name }}
+    app.kubernetes.io/managed-by: summon-operator
+    {{/* creator and source needed for goldilocks dashboard to pick vpa up */ -}}
+    creator: "Fairwinds"
+    source: "goldilocks"
+spec:
+  targetRef: {{ block "controller" . }}{{ end }}
+  updatePolicy:
+    updateMode: {{ block "updateMode" . }}{{ end }}
```
coderanger, line 19: If you do `{{ block "updateMode" . }}Off{{ end }}` then you don't need to put the default in each other template :) It serves as a default value for the block.

PR #269, opened 2020-04-02
🐛 Add a quick hack for DISPATCH_BASE_URL setting to only exist if the…

coderanger:
… component is on.

This was a misunderstanding with the comp-dispatch team. I've added a DISPATCH_ENABLED boolean flag for a better solution for the future, but for now make it conditional so we can silence errors in PCR-4/5 (i.e. until a fixed Summon is available).

### DIFF
--- a/pkg/controller/summon/components/defaults_test.go
+++ b/pkg/controller/summon/components/defaults_test.go
@@ -242,7 +242,7 @@ var _ = Describe("SummonPlatform Defaults Component", func() {
 		instance.Spec.Dispatch.Version = "foo"
 		Expect(comp).To(ReconcileContext(ctx))
 		Expect(instance.Spec.Config["DISPATCH_ENABLED"].Bool).To(PointTo(BeTrue()))
-		Expect(instance.Spec.Config["DISPATCH_BASE_URL"].String).To(Equal("http://foo-dev-dispatch:8000/"))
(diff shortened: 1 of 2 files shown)


pkg/controller/summon/components/defaults_test.go
```diff
@@ -237,4 +237,18 @@ var _ = Describe("SummonPlatform Defaults Component", func() {
 			Expect(instance.Spec.Replicas.CeleryBeat).To(PointTo(BeEquivalentTo(1)))
 		})
 	})
+
+	It("sets DISPATCH_ENABLED if the dispatch component is enabled", func() {
+		instance.Spec.Dispatch.Version = "foo"
+		Expect(comp).To(ReconcileContext(ctx))
+		Expect(instance.Spec.Config["DISPATCH_ENABLED"].Bool).To(PointTo(BeTrue()))
```
vivekjainx86, line 244: A small doubt here, In the code, DISPATCH_ENABLE flag is depend on only Replicas.Dispatch value. So, how it will be TRUE by Dispatch.Version value?

coderanger: Because in replicaDefaults down below, we set the default replicas to 1 if a version is set for the component, and 0 if it isn't :)

vivekjainx86: Oh ok, I got it now.

PR #272, opened 2020-04-15
Horizontal pod autoscaling

rc-marina:
TODOS in future PR: 
- ~~finalizer logic to clean up HPAs when autoscaling disabled~~
- complete implementation around remaining components that need Spec.Replicas.*Auto (and probably merge functions `NewDeployment()` `NewDeploymentWithAutoscaling()`)
- SummonPlatform Spec fields for setting min/max replicas on HPA autoscalers and set defaults 
based on environment (dev, qa, uat, prod) -- if <component>Auto used, have the respective `.Instance.Spec.Replicas.<component>` set the default minReplica for HPA?
- increase default maxReplicas from 10 to 32?


pkg/controller/summon/components/defaults_test.go
```diff
@@ -128,6 +128,13 @@ var _ = Describe("SummonPlatform Defaults Component", func() {
 		Expect(instance.Spec.Replicas.CeleryBeat).To(PointTo(BeEquivalentTo(1)))
 	})
 
+	It("sets default component autoscaling to false", func() {
+		instance.Spec.Version = "1.2.3"
+		Expect(comp).To(ReconcileContext(ctx))
+		Expect(instance.Spec.Replicas.CelerydAuto).To(PointTo(BeEquivalentTo(false)))
+		// Expect(instance.Spec.Replicas.Web).To(PointTo(BeEquivalentTo(false)))
```
coderanger, line 135: Line to clean up :)

pkg/controller/summon/components/hpa_test.go
```diff
@@ -0,0 +1,103 @@
...
+			comp = summoncomponents.NewHPA("celeryd/hpa.yml.tpl", func(s *summonv1beta1.SummonPlatform) bool { return *s.Spec.Replicas.CelerydAuto })
+			Expect(comp).To(ReconcileContext(ctx))
+			hpa := &autoscalingv2beta2.HorizontalPodAutoscaler{}
+			err := ctx.Client.Get(context.TODO(), types.NamespacedName{Name: "foo-dev-celeryd-hpa", Namespace: instance.Namespace}, hpa)
+			Expect(err).To(HaveOccurred())
+		})
+	})
+
+	/* TODO: Test finalizer logic
+	Context("when ReplicaSpecs.<component>Auto was true, but set to false", func() {
```
coderanger, line 91: So the way to do this is to put the HPA object into the fake client:

```
hpa := &autoscalingv2beta2{... some stuff here ...}
ctx.Client = fake.NewFakeClient(instance, hpa)
```

That will make it look like the HPA object already existed when the test starts.

pkg/controller/summon/templates/celeryd/hpa.yml.tpl
```diff
@@ -0,0 +1,15 @@
...
+{{ define "componentType" }}worker{{ end }}
+{{ define "target"}}
+    apiVersion: "apps/v1"
+    kind: Deployment
+    name: {{ .Instance.Name }}-celeryd{{ end }}
+{{ define "metric" }}
+        name: ridecell:rabbitmq_summon_celery_queue_scaler
+        selector:
+          matchLabels: 
+            vhost: {{ .Instance.Name | quote }}{{ end }}
```
coderanger, line 11: If you do `{{- end }}` you can put it on its own line. The `-` removes the extra whitespace.

pkg/controller/summon/templates/celeryd/hpa.yml.tpl
```diff
@@ -0,0 +1,15 @@
...
+{{ define "mTarget" }}
+        type: Value
+        value: 1{{ end }}
+{{ template "hpa" . }}
```
coderanger, line 15: Possibly helpful thing, we could override the `minReplicas` block and use the value of `.Instance.Spec.Replics.Celeryd` as the minimum. Thoughts?

pkg/controller/summon/templates/helpers/hpa.yml.tpl
```diff
@@ -0,0 +1,23 @@
...
+    app.kubernetes.io/name: {{ block "componentName" . }}{{ end }}
+    app.kubernetes.io/instance: {{ .Instance.Name }}-{{ block "componentName" . }}{{ end }}
+    app.kubernetes.io/version: {{ .Instance.Spec.Version }}
+    app.kubernetes.io/component: {{ block "componentType" . }}{{ end }}
+    app.kubernetes.io/part-of: {{ .Instance.Name }}
+    app.kubernetes.io/managed-by: summon-operator
+spec:
+  scaleTargetRef: {{ block "target" . }}{{ end }}
+  minReplicas: {{ block "minReplicas" . }}1{{ end }}
+  maxReplicas: {{ block "maxReplicas" . }}10{{ end }}
```
coderanger, line 17: Might need more eventually since we've had up to like 32 in prod before, but let's worry about that once the whole thing works :D

PR #274, opened 2020-04-24
[WIP] refactor how we configure Autoscaling and allow setting of min/max replicas

rc-marina:
Might add on to this PR with updated HPA templates for the other summon components (web, static, hwAux, etc) if prometheus metrics for them become available.

### DIFF
--- a/pkg/apis/summon/v1beta1/summonplatform_types.go
+++ b/pkg/apis/summon/v1beta1/summonplatform_types.go
@@ -104,6 +104,17 @@ type MigrationOverridesSpec struct {
 	RedisHostname     string `json:"redisHostname,omitempty"`
 }
 
+type AutoConfigSpec struct {
+	// Control if Horizontal Pod Autoscaling is used to manage replica counts.
+	HpaEnabled *bool `json:"enabled,omitempty"`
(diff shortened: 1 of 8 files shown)


pkg/apis/summon/v1beta1/summonplatform_types.go
```diff
@@ -104,6 +104,17 @@ type MigrationOverridesSpec struct {
 	RedisHostname     string `json:"redisHostname,omitempty"`
 }
 
+type AutoConfigSpec struct {
+	// Control if Horizontal Pod Autoscaling is used to manage replica counts.
+	HpaEnabled *bool `json:"enabled,omitempty"`
```
coderanger, line 109: I would keep the JSON and Struct matching so either should both be `HpaEnabled/hpaEnabled` or `Enabled/enabled`. Makes it easier to go back and forth mentally :) I would probably favor `Enabled` as simpler but will we regret that later?

rc-marina: keeping it as HpaEnabled to allow for other possible future enablement flags (e.g vpaEnabled config?)

PR #284, opened 2020-05-19
CORE-1558 Enable JSON Logging for Dev, QA, UAT envs

ryfloyd:


rc-marina:
This is a temporary fix to unbreak SumoLogic Dashboards used by support team.

The proposed plan:
1. ridecell-operator defaults to TRUE for all but prod.
2. `ENABLE_JSON_LOGGING` to be set to `True` on Prod instances during future deploys (making sure Customer Support team is aware) 
3. A PR to ridecell/summon-platform will be made to remove `ENABLE_JSON_LOGGING` from settings. Turning it on everywhere and without even checking config values. So no turning it off.
4. A follow-up PR to ridecell-operator to clean and remove `ENABLE_JSON_LOGGING`





2:18
4. We remove the references to ENABLE_JSON_LOGGING from ridecell-operator and from kuberenetes-summon

PR #294, opened 2020-06-10
HW-2318 devops config for optimus

colonoh:
jira ref: https://ridecell.atlassian.net/browse/HW-2318

### DIFF
--- a/pkg/controller/summon/templates/aws/iamuser.yml.tpl
+++ b/pkg/controller/summon/templates/aws/iamuser.yml.tpl
@@ -29,6 +29,21 @@ spec:
                 }
               ]
             }
+{{if .Extra.lgBucketName}}
+   allow_s3_lg: |
+            {
+               "Version": "2012-10-17",
+               "Statement": [
+                {
+                   "Effect": "Allow",
+                   "Action": [
+                      "s3:GetObject",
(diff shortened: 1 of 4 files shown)


pkg/controller/summon/templates/aws/iamuser.yml.tpl
```diff
@@ -29,6 +29,21 @@ spec:
...
             }
+{{if .Extra.optimusBucketName}}
+   allow_s3_optimus: |
+            {
+               "Version": "2012-10-17",
+               "Statement": [
+                {
+                   "Effect": "Allow",
+                   "Action": [
+                      "s3:GetObject"
```
vivekjainx86, line 40: don't we require ListObject perm too? or its sufficient?

colonoh: @personalcomputer and I tested and `GetObject` is sufficient.  We actually do not want `ListObject` so that it is not possible for us to list the files.

colonoh: @vivekjainx86 For some context, we have an API which provides the bucket's object key to us so we don't need (or want) to list all the files.

PR #300, opened 2020-07-29
CORE-1558 Enable JSON Logging on all Production ENVs

ryfloyd:
Related to #284 

The proposed plan:

1. ridecell-operator defaults to TRUE for all but prod.
2. ENABLE_JSON_LOGGING to be set to True on Prod instances during future deploys (making sure Customer Support team is aware)
3. A PR to ridecell/summon-platform will be made to remove ENABLE_JSON_LOGGING from settings. Turning it on everywhere and without even checking config values. So no turning it off.
4. A follow-up PR to ridecell-operator to clean and remove ENABLE_JSON_LOGGING
5. We remove the references to ENABLE_JSON_LOGGING from ridecell-operator and from kuberenetes-summon

### DIFF
--- a/pkg/controller/summon/components/defaults.go
+++ b/pkg/controller/summon/components/defaults.go
@@ -262,9 +262,8 @@ func (comp *defaultsComponent) Reconcile(ctx *components.ComponentContext) (comp
 		gatewayEnv = "master"
 	}
 
-	if instance.Spec.Environment != "prod" {
-		defBoolVal("ENABLE_JSON_LOGGING", true)
-	}
+	// Set JSON Logging globally if not already set.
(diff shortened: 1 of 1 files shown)


avdhoot:
@ryfloyd let us know when you want this change in prod.

ryfloyd:
@avdhoot Yes, CS team is aware. We do have a few alerts configured ... but don't believe they are working?  https://github.com/Ridecell/ridecell-alerts/blob/master/aws-eu-prod/carsharing-alerts.yml#L17 But that search query for both the non-json and json formatting.

I'd like for this to be deployed along side next PCR deploy.

PR #304, opened 2020-08-31
DEVOPS-2461 Adding iamrole support for summonplatform

vivekjainx86:
Added IAMRole support for Summon platform, for testing, we can enable use of IAMRole instead of IAMUser by following Summon Spec.
`useIamRole: true`

### DIFF
--- a/pkg/controller/summon/templates/helpers/deployment.yml.tpl
+++ b/pkg/controller/summon/templates/helpers/deployment.yml.tpl
@@ -34,6 +34,9 @@ spec:
         iam.amazonaws.com/role: summon-platform-{{ .Instance.Spec.Environment }}-{{ .Instance.Name }}
         {{ end }}
     spec:
+      {{ if .Instance.Spec.UseIamRole }}
+      serviceAccountName: {{ .Instance.Name }}
+      {{ end }}
       affinity:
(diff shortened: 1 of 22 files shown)


avdhoot (dismissed):

avdhoot (approved):

avdhoot (approved):

pkg/controller/summon/templates/helpers/deployment.yml.tpl
```diff
@@ -30,7 +30,13 @@ spec:
       annotations:
         summon.ridecell.io/appSecretsHash: {{ .Extra.appSecretsHash }}
         summon.ridecell.io/configHash: {{ .Extra.configHash }}
+        {{ if .Instance.Spec.UseIamRole }}
+        iam.amazonaws.com/role: summon-platform-{{ .Instance.Spec.Environment }}-{{ .Instance.Name }}
```
vivekjainx86, line 34: @avdhoot do we need to keep these annotations now?

avdhoot: let's keep for now. We will remove it when we sure that's service account based IAM role work.

PR #310, opened 2020-10-06
DEVOPS-2461 Web identity 2

avdhoot:


pkg/controller/summon/components/service_account_k8s.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if !instance.Spec.UseIamRole {
+		return components.Result{}, nil
+	}
+	//TODO Add better way to get accountID
+	permissionsBoundaryArn := os.Getenv("PERMISSIONS_BOUNDARY_ARN")
+	if permissionsBoundaryArn == "" {
+		return components.Result{}, errors.Errorf("serviceaccountk8s: permissions_boundary_arn is empty")
+	}
+	match := regexp.MustCompile(`:([0-9]{6,}):`).FindStringSubmatch(permissionsBoundaryArn)
```
vivekjainx86, line 59: The report controller will also require aws account id, can we make it an ENV var?

avdhoot: That's plan but not part of this PR. Please check `todo` on line no 54

vivekjainx86: In fact, I already used AWS_ACCOUNT_ID env var in report controller code.

vivekjainx86: ohh, ok, we can modify this later.

vivekjainx86 (approved):
looks good

pkg/controller/summon/components/service_account_k8s.go
avdhoot: We are using AWS_ACCOUNT_ID multiple places. New PR will fix it everywhere.

PR #325, opened 2020-12-23
some changes

vivekjainx86:
deployment,yml.tpl
- removed metrics spec
- removed default value for kafka consumer
- removed Zone failure pod affinity
- modified cpu/memory resources
summonplatform_types.go
- removed metricsspec related change


avdhoot:
we can scale replica if needed. Hence reduced resource limit as per current resource utilization.

AssafRideCell (approved):
LGTM

